PR #4, opened 2017-08-17
View fees table feature

rattur:
### JIRA link ###

https://tools.hmcts.net/jira/browse/ROC-966

### Change description ###

Fees table

```
[ ] Yes
[x] No
```


src/main/app/fees/moneyConvertor.ts
```diff
@@ -0,0 +1,10 @@
+export default class MoneyConvertor {
```
rlewan: Should be `Converter`, not `Convertor`. Otherwise looks good :)

PR #7, opened 2017-08-18
Add testing support feature

timja-kainos:


src/main/app/claims/claimStoreClient.ts
```diff
@@ -158,4 +160,18 @@ export default class ClaimStoreClient {
       }
     })
   }
+
+  static updateResponseDeadline (updateRequest: UpdateResponseDeadlineRequest, user: User) {
```
rlewan: Not sure if this should be in claim store client if it's testing-support related. Maybe this could be another client, what do you think?

timja-kainos: done and also fixed some boolean issues with the feature toggle

PR #18, opened 2017-08-23
Bug: fixed total amount calculations

yarpo:
### Change description ###
Fixed calculation of claim total amount.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/app/claims/models/claim.ts
+++ b/src/main/app/claims/models/claim.ts
@@ -44,9 +44,10 @@ export default class Claim implements Serializable<Claim> {
     const interestRate = this.claimData.interest
     const interestDate = this.claimData.interestDate
     const claimAmount: number = this.claimData.amount
-    const date = interestDate.type === InterestDateType.SUBMISSION ? this.createdAt : interestDate.date
+    const paidFeeAmount: number = this.claimData.paidFeeAmount || 0
+    const date = (interestDate.type === InterestDateType.SUBMISSION) ? this.createdAt : interestDate.date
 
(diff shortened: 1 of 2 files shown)


src/main/app/claims/models/claim.ts
```diff
@@ -44,9 +44,10 @@ export default class Claim implements Serializable<Claim> {
     const interestRate = this.claimData.interest
     const interestDate = this.claimData.interestDate
     const claimAmount: number = this.claimData.amount
-    const date = interestDate.type === InterestDateType.SUBMISSION ? this.createdAt : interestDate.date
+    const paidFeeAmount: number = this.claimData.paidFeeAmount || 0
+    const date = (interestDate.type === InterestDateType.SUBMISSION) ? this.createdAt : interestDate.date
```
timja-kainos: no need for brackets here

src/main/app/claims/models/claim.ts
```diff
@@ -44,9 +44,10 @@ export default class Claim implements Serializable<Claim> {
     const interestRate = this.claimData.interest
     const interestDate = this.claimData.interestDate
     const claimAmount: number = this.claimData.amount
-    const date = interestDate.type === InterestDateType.SUBMISSION ? this.createdAt : interestDate.date
+    const paidFeeAmount: number = this.claimData.paidFeeAmount || 0
```
timja-kainos: not the right fix

src/main/app/claims/models/claim.ts
yarpo: I think it's much more readable

src/main/app/claims/models/claim.ts
yarpo: ok

PR #24, opened 2017-08-25
Adjust T1 frontend to claim store change

rlewan:
### JIRA link ###

https://tools.hmcts.net/jira/browse/ROC-2179

### Change description ###

Updates for T1 frontend to work with unified claim store model.

```
[x] Yes
[ ] No
```


src/main/app/claims/claimModelConverter.ts
```diff
@@ -14,12 +14,15 @@ export class ClaimModelConverter {
...
 
-    draftClaim.claimant.dateOfBirth = draftClaim.claimant.dateOfBirth.date.asString() as any
     draftClaim.reason = draftClaim.reason.reason as any
 
+    draftClaim['feeAmountInPennies'] = draftClaim.claimant.payment.amount
+
     draftClaim['payment'] = draftClaim.claimant.payment as any
     delete draftClaim.claimant.payment
 
+    draftClaim.amount['type'] = 'breakdown'
```
yarpo: don't we have any enum for that?

src/main/app/claims/claimModelConverter.ts
```diff
@@ -36,6 +39,8 @@ export class ClaimModelConverter {
     draftClaim.claimant['type'] = 'individual'
 
     draftClaim.claimant.mobilePhone = draftClaim.claimant.mobilePhone.number as any
+
+    draftClaim.claimant.dateOfBirth = draftClaim.claimant.dateOfBirth.date.asString() as any
```
yarpo: don't we always want to keep dates as Moment?

src/main/app/claims/claimModelConverter.ts
```diff
@@ -58,6 +63,9 @@ export class ClaimModelConverter {
     }
 
     draftClaim.defendant['type'] = 'individual'
```
yarpo: enum?

src/main/app/claims/claimModelConverter.ts
timja-kainos: nope its not an enum, easier to just hardcode it here I would say, we shouldn't ever have to look at it.

src/main/app/claims/claimModelConverter.ts
timja-kainos: this function converts to moment and then formats it how as want it, asString() may not be the best name, but it is what we want to do here
```
this.toMoment().format(DATE_FORMAT)
```

src/main/app/claims/claimModelConverter.ts
timja-kainos: anirudha's PR has this completely reworked, this is temporary

src/main/app/claims/models/claimData.ts
```diff
@@ -8,18 +8,30 @@ import Payment from 'app/pay/payment'
...
+      throw new Error('This claim has multiple defendants')
+    }
+  }
+
   deserialize (input: any): ClaimData {
     if (input) {
       this.claimant = new Claimant().deserialize(input.claimant)
-      this.defendant = new Defendant().deserialize(input.defendant)
+      let defendants: Defendant[] = []
+      input.defendants.forEach((defendant: Defendant) => {
```
timja-kainos: this should be a map

PR #28, opened 2017-08-30
ROC-1532: CCJ payment options

yarpo:
### JIRA link ###
https://tools.hmcts.net/jira/browse/ROC-1532

### Change description ###
Added page to select type of payment

```
[ ] Yes
[x] No
```


src/main/features/ccj/routes/payment-options.ts
```diff
@@ -0,0 +1,48 @@
...
+              res.redirect(Paths.checkYourAnswerPage.uri.replace(':externalId', externalId))
+              break
+            case PaymentType.FULL:
+              res.redirect(Paths.payBySetDatePage.uri.replace(':externalId', externalId))
+              break
+            case PaymentType.BY_INSTALMENTS:
+              res.redirect(Paths.repaymentPlanPage.uri.replace(':externalId', externalId))
+              break
+            default:
+              console.log(form.model)
```
robertparkinson: I don't think we should be leaving this in here?

yarpo: well spotted :)

yarpo: removed

PR #32, opened 2017-09-01
Method to merge path URI with path parameters

geek-soft:
### JIRA link ###

No

### Change description ###

A method to evaluate URI from the template with placeholders and an array of path parameters.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/common/router/routablePath.ts
+++ b/src/main/common/router/routablePath.ts
@@ -34,7 +34,7 @@ export class RoutablePath {
   }
 
   evaluateUri (substitutions: {[key: string]: string}): string {
-    if (substitutions == null || Object.keys(substitutions).length === 0) {
+    if (substitutions === undefined || Object.keys(substitutions).length === 0) {
       throw new Error('Path parameter substitutions are required')
     }
 
--- a/src/main/common/router/routablePath.ts
+++ b/src/main/common/router/routablePath.ts
@@ -33,18 +33,25 @@ export class RoutablePath {
     return `${featureName}/views/${viewPath}`
   }
 
-  evaluateUri (substitutions: {[key: string]: string}): string {
+  evaluateUri (substitutions: { [key: string]: string }): string {
     if (substitutions === undefined || Object.keys(substitutions).length === 0) {
       throw new Error('Path parameter substitutions are required')
     }
 
     const path = Object.entries(substitutions).reduce((uri: string, substitution: [string, string]) => {
       const [parameterName, parameterValue] = substitution
-      return uri.replace(`:${parameterName}`, parameterValue)
+
+      const updatedUri: string = uri.replace(`:${parameterName}`, parameterValue)
+      if (updatedUri === uri) {
+        throw new Error(`Path parameter :${parameterName} is not defined`)
+      }
+      return updatedUri
     }, this.uri)
 
-    if (pathParameterRegex.test(path)) {
-      throw new Error('At least one path parameter substitution is missing')
+    const missingParameters = path.match(pathParameterRegex)
+    if (missingParameters) {
+      const removeLeadingSlash = value => value.substring(1)
+      throw new Error(`Path parameter substitutions for ${missingParameters.map(removeLeadingSlash).join(', ')} are missing`)
     }
 
     return path
--- a/src/main/common/router/routablePath.ts
+++ b/src/main/common/router/routablePath.ts
@@ -1,3 +1,5 @@
+const pathParameterRegex = /\/:[^\/]+/g
+
 export class RoutablePath {
   _uri: string
 
@@ -16,12 +18,12 @@ export class RoutablePath {
   get associatedView (): string {
     if (!this.feature) {
       return this._uri
-        .replace(/\/:[^\/]+/g, '') // remove path params
+        .replace(pathParameterRegex, '') // remove path params
         .substring(1) // remove leading slash
     }
 
     const split: string[] = this._uri
-      .replace(/\/:[^\/]+/g, '') // remove path params
+      .replace(pathParameterRegex, '') // remove path params
       .substring(1) // remove leading slash
       .split('/')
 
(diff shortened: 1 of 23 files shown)


src/main/common/router/routablePath.ts
```diff
@@ -30,4 +32,21 @@ export class RoutablePath {
     const viewPath: string = split.slice(isCaseUri ? 2 : 1).join('/')
     return `${featureName}/views/${viewPath}`
   }
+
+  evaluateUri (substitutions: {[key: string]: string}): string {
+    if (substitutions == null || Object.keys(substitutions).length === 0) {
```
timja-kainos: !substitutions? we don't use null in our codebase??

geek-soft: I made that deliberately thinking of this as something we could to extract in the future. I tend to use nulls only for places which are generic enough to be extracted in the future.

timja-kainos: even so other people shouldn't use null either, even if extracted it should still be the same logic I would have thought?

geek-soft: Agree, that is why I decided to check for null as well. In first version I only had `substitutions  === undefined` but that would allow nulls for those who did not discard them.

A `substitutions == null` is doing both `substitutions === undefined` and `substitutions === null`. I did not want to use `!substitutions` since the that covers too many cases including `false` or `''`.

src/test/a11y/a11y.ts
```diff
@@ -68,7 +68,7 @@ describe('Accessibility', () => {
     Object.values(pathsRegistry).forEach((path: RoutablePath) => {
       const excluded = excludedPaths.some(_ => _ === path)
       if (!excluded) {
-        check(path.uri)
+        check(path.evaluateUri({ externalId: '91e1c70f-7d2c-4c1e-a88f-cbb02c0e64d6' }))
```
yarpo: do we need it?

yarpo: I mean `externalId` is not used anywhere as we mock methods with no arguments description

src/test/common/router/routablePath.ts
```diff
@@ -12,6 +12,30 @@ describe('RoutablePath', () => {
...
+        .to.throw(Error, 'At least one path parameter substitution is missing')
+    })
+
+    it('should replace all path parameter placeholders', () => {
+      expect(new RoutablePath('/case/:id/payment/:payment-type').evaluateUri({ id: '999', 'payment-type': 'card' }))
+        .to.be.equal('/case/999/payment/card')
+    })
+
+    it('should ignore substitutions which are not defined as path parameter placeholders', () => {
+      expect(new RoutablePath('/case/:id').evaluateUri({ id: '999', foo: 'bar' }))
```
yarpo: Why don't we throw exception for unknown params given?

src/test/features/dashboard/routes/claimant.ts
```diff
@@ -64,8 +64,7 @@ describe('Dashboard - claimant page', () => {
         await request(app)
           .post(claimantPage)
           .set('Cookie', `${cookieName}=ABC`)
-          .expect(res => expect(res).to.be.redirect.toLocation(
-            CCJPaths.theirDetailsPage.uri.replace(':externalId', sampleClaimDraftObj.externalId)))
+          .expect(res => expect(res).to.be.redirect.toLocation(CCJPaths.theirDetailsPage.evaluateUri({ externalId: sampleClaimDraftObj.externalId })))
```
yarpo: is this line shorter than 120 chars?

src/test/a11y/a11y.ts
geek-soft: Used in one view where nextUrl is validated.

src/test/common/router/routablePath.ts
geek-soft: I had this though but gave up to keep it simpler. But generally it would be good to be stricter.

src/test/features/dashboard/routes/claimant.ts
geek-soft: Not sure, I tend to ignore max length restrictions as I think breaking when line break is forced by such constraint it does not make code more readable.

timja-kainos: 1. it makes reviewing harder
2. it makes seeing it in the IDE harder.

I think its worth breaking it on `toLocation`

rattur: Could have defined sampleClaimDraftObj.externalId as a const but otherwise it looks fine

src/test/common/router/routablePath.ts
timja-kainos: is it more complicated than looking for `/:foo/` if so leave it,
otherwise it would be nice to have

src/main/common/router/routablePath.ts
```diff
@@ -30,4 +32,21 @@ export class RoutablePath {
...
   }
+
+  evaluateUri (substitutions: {[key: string]: string}): string {
+    if (substitutions === undefined || Object.keys(substitutions).length === 0) {
+      throw new Error('Path parameter substitutions are required')
+    }
+
+    const path = Object.entries(substitutions).reduce((uri: string, substitution: [string, string]) => {
+      const [parameterName, parameterValue] = substitution
+      return uri.replace(`:${parameterName}`, parameterValue)
```
timja-kainos: Would this work?
```
const replacedValue = uri.replace(`:${parameterName}`, parameterValue)
return replacedValue !== uri ? replacedValue : throw new Error('Didn’t find match')
```

timja-kainos: Some other options on here:
https://stackoverflow.com/questions/5257000/how-to-know-if-javascript-string-replace-did-anything

src/test/features/dashboard/routes/claimant.ts
timja-kainos: @rattur ?

src/main/common/router/routablePath.ts
geek-soft: Logically should work.

geek-soft: Done it this way. Not exactly the syntax but logic the same

src/test/features/dashboard/routes/claimant.ts
geek-soft: Reverted line break

PR #34, opened 2017-09-04
Add feature toggle for CCJ

timja-kainos:


src/main/app/claims/models/claim.ts
```diff
@@ -55,6 +57,10 @@ export default class Claim implements Serializable<Claim> {
   }
 
   get eligibleForCCJ (): boolean {
+    if (!toBoolean(config.get<boolean>('featureToggles.countyCourtJudgment'))) {
```
yarpo: have you covered it in unit tests?

timja-kainos: no

PR #43, opened 2017-09-11
[DO NOT MERGE] FIX: defendant email mapping draft -> claim

yarpo:
### Change description ###
This is bug fix for production.
When (optional) defendant email is not provided our converter was sending invalid data structure to claim API (string expected, object given) that caused 500 internal error on API side. 

It's already fixed in the newest version of cmc frontend. This fix is a temporary solution only for currently deployed version of cmc citizen frontend.

The same scenario was tested on the latest version of application and everything works well.

```
[x] Yes
[ ] No
```


geek-soft:
Not gonna be merged. New branch created.

PR #47, opened 2017-09-12
Bump version for future development

geek-soft:
### JIRA link ###

None

### Change description ###

Version bump for future development

```
[ ] Yes
[x] No
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "citizen-frontend",
-  "version": "2.1.0",
+  "version": "1.1.0",
   "private": true,
   "engines": {
     "node": ">=7.10.1"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "citizen-frontend",
-  "version": "1.0.1",
+  "version": "2.1.0",
   "private": true,
   "engines": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "citizen-frontend",
-  "version": "1.0.1",
+  "version": "2.1.0",
```
timja-kainos: I wouldn't want to tie us to JIRA versions it will never work...
maybe 2.0.0 here, or the next minor in 1.x

geek-soft: I though maybe to make JIRA and semver in sync for citizen-frontend & claim-store. Does not apply for other components really. That could work for major & minor releases since usually both get upgraded. That still might be out of sync for hotfix releases hoverer that could still work, lets say we take 2.1.0 of claim store and 2.1.1 of citizen frontend.

I do not feel strong about this, just wanted to kick off this discussion.

timja-kainos: problem is we would already have bumped the version for 2.1.0 to go out the door
but then turns out the next release is going to be 2.0.1 because its a hotfix.

I really don't think we should be tying internal numbers to external numbers.

timja-kainos:
Merged to fix dev

PR #50, opened 2017-09-13
Use apostrophe instead of single quote char

geek-soft:
### JIRA link ###

None

### Change description ###

Change introduces use of apostrophe instead of single quote char + one word change at Ben's request

```
[ ] Yes
[x] No
```


geek-soft:
When that PR is approved, I will do appropriate change to IT to make build pass.

src/test/features/claim/routes/fees.ts
```diff
@@ -89,7 +89,7 @@ describe('Claim issue: fees page', () => {
         await request(app)
           .get(ClaimPaths.feesPage.uri)
           .set('Cookie', `${cookieName}=ABC`)
-          .expect(res => expect(res).to.be.successful.withText('Fees you&#39;ll pay'))
```
timja-kainos: <3 tried to stop this going in but my comment got lost when a change happened

PR #51, opened 2017-09-13
Do not display defendant date of birth page when CCJ is being requested against non individual defendants

geek-soft:
### JIRA link ###

https://tools.hmcts.net/jira/browse/ROC-1531

### Change description ###

This is to hide defendant date of birth page when CCJ is requested for claim issued against non individual defendant. 

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/views/includes/macros/partyDetails.njk
+++ b/src/main/views/includes/macros/partyDetails.njk
@@ -74,7 +74,7 @@
     }}
   {% endif %}
 
-  {% if (partyDetails.type === 'individual' and partyDetails.dateOfBirth) %}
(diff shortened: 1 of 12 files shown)


src/main/views/includes/macros/partyDetails.njk
```diff
@@ -74,7 +74,7 @@
     }}
   {% endif %}
 
-  {% if (partyDetails.type === 'individual' and partyDetails.dateOfBirth) %}
+  {% if (partyDetails.type === 'individual' and partyDetails.dateOfBirth and partyDetails.dateOfBirth.known ) %}
```
geek-soft: Need to ask @BenClancy what to do if DOB is not known. It hides it for now however we could say `Not sure` for example.

PR #52, opened 2017-09-14
Use SonarQube for code analysis

geek-soft:
### JIRA link ###

None

### Change description ###

Introduces SonarQube to our frontend project.

```
[ ] Yes
[x] No
```


README.md
```diff
@@ -49,16 +49,31 @@ Running the linting:
...
+
+```bash
+$ yarn test:a11y
+```
+
+For test coverage:
+
+```bash
+$ yarn test:coverage
+```
```
timja-kainos: probably worth adding this section for now to the README:
https://github.com/hmcts/cmc-legal-rep-frontend#warnings-while-running-yarn-install-on-yarn-version-101

geek-soft: Saw the warning but it does not break anything. My approach would be the same as regarding other warnings, I mean ignore ;) I do not like the fact somebody has to remove it at some point.

```
warning "request-promise-native@1.0.4" has incorrect peer dependency "request@^2.34".
warning "request-promise-core@1.1.1" has incorrect peer dependency "request@^2.34".
warning "react-dom@15.6.1" has incorrect peer dependency "react@^15.6.1".
warning "ajv-keywords@1.5.1" has incorrect peer dependency "ajv@>=4.10.0".
```

timja-kainos: but people will waste time asking about it and trying to track it down, better to have something to point to

geek-soft: When people see the problem, they either won't care or won't look into README page to find the answer, sooner they will search Google.

Regardless the fact I disagree with the principle I will add this as there is no harm and I want to make you happy ;)

.travis.yml
```diff
@@ -3,12 +3,13 @@ cache: yarn
 script:
   - yarn --version
   - yarn setup
-  - yarn check
+  - yarn check --integrity
   - yarn lint
   - yarn test
```
timja-kainos: do we still need to run test and test:routes if coverage is doing it?

geek-soft: Possibly not on travis.

Left `Test (unit)` and `Test (routes)` stages on Jenkins to have it visibly separated of what has failed especially that redoing this as a part of coverage gives just few extra secs to the run time, however I do not feel very strong about that.

PR #58, opened 2017-09-15
Send custom GA events on defendant DOB selection or payment option selection

geek-soft:
### JIRA link ###

https://tools.hmcts.net/jira/browse/ROC-2343

### Change description ###

This story records both:
- defendant date of birth selection
- payment option selection

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/features/ccj/views/date-of-birth.njk
+++ b/src/main/features/ccj/views/date-of-birth.njk
@@ -7,7 +7,7 @@
 {% block content %}
   <div class="grid-row">
     <div class="column-two-thirds">
-      <form method="post" novalidate>
+      <form method="post" class="analytics-submit-event-trigger" data-event-action="CCJ: Select defendant date of birth" data-event-label-from="known" novalidate>
         {{ csrfProtection(csrf) }}
 
         <h1 class="heading-large">
--- a/src/main/public/js/analytics-custom-event-sender.js
+++ b/src/main/public/js/analytics-custom-event-sender.js
@@ -10,9 +10,7 @@ $(function () {
       throw new Error('Event label is required')
     }
 
-    if (eventAction && eventLabel) {
-      ga('send', 'event', eventCategory, eventAction, eventLabel)
-    }
+    ga('send', 'event', eventCategory, eventAction, eventLabel)
   }
 
   function labelText(labelElement) {
--- a/src/main/public/js/analytics-custom-event-sender.js
+++ b/src/main/public/js/analytics-custom-event-sender.js
@@ -27,15 +27,15 @@ $(function () {
     return undefined
   }
 
-  function formValue (formElement, inputName) {
-    if (!formElement) {
-      throw new Error('Form element is required')
+  function formValue (form, inputName) {
+    if (!form) {
+      throw new Error('Form is required')
     }
     if (!inputName) {
       throw new Error('Input name is required')
     }
 
-    var inputElement = formElement.find('input[name=' + inputName + ']')
+    var inputElement = form.find('input[name=' + inputName + ']')
     if (inputElement.length > 0) {
       switch (inputElement[0].type) {
         case 'radio':
@@ -43,7 +43,7 @@ $(function () {
             return undefined
           }
 
-          return labelText(formElement.find('label[for=' + inputElement.filter(':checked')[0].id + ']'))
+          return labelText(form.find('label[for=' + inputElement.filter(':checked')[0].id + ']'))
         default:
           throw new Error('Input type is not supported')
       }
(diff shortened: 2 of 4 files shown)


geek-soft:
screen shot 2017-09-15 at 20 09 12

Events are visible in the real time view. Labels saying what value is submitted is only available next day.

src/main/public/js/analytics-custom-event-sender.js
```diff
@@ -0,0 +1,69 @@
...
+      throw new Error('Event action is required')
+    }
+    if (!eventLabel) {
+      throw new Error('Event label is required')
+    }
+
+    ga('send', 'event', eventCategory, eventAction, eventLabel)
+  }
+
+  function labelText(labelElement) {
```
timja-kainos: I wonder whether the label text should be going into the action, https://support.google.com/analytics/answer/1033068?hl=en

Real time view only goes back 30 minutes for me,
I found the other data in Behaviour => Events, but I can't find a way to tie the label back to an action.
And 'Not sure' is a bit too generic not to be able to tie it back.

src/main/public/js/analytics-custom-event-sender.js
```diff
@@ -0,0 +1,69 @@
...
+    for (var content of labelElement.contents()) {
+      const text = $(content).text().trim()
+      if (text && content.class !== 'form-hint') {
+        return text
+      }
+    }
+    return undefined
+  }
+
+  function formValue(formElement, inputName) {
```
rlewan: I think this name is a little unfortunate because it ultimately retrieves a label, not a value.

rlewan: I think I'd also use just `form` for first parameter (and update in other places accordingly). This confused me a little because I thought it's an element of a form, not the form itself. Not strong on it though.

src/main/public/js/analytics-custom-event-sender.js
```diff
@@ -0,0 +1,69 @@
...
+    if (!inputName) {
+      throw new Error('Input name is required')
+    }
+
+    var inputElement = formElement.find('input[name=' + inputName + ']')
+    if (inputElement.length > 0) {
+      switch (inputElement[0].type) {
+        case 'radio':
+          if (!inputElement.is(':checked')) {
+            return undefined
```
rlewan: This will propagate up to `sendEvent` function and write errors in JavaScript console. Nut a big problem but not too elegant as well I think.

src/main/features/ccj/views/date-of-birth.njk
```diff
@@ -7,7 +7,7 @@
 {% block content %}
   <div class="grid-row">
     <div class="column-two-thirds">
-      <form method="post" novalidate>
+      <form method="post" class="analytics-submit-event-trigger" data-event-action="CCJ: Select defendant date of birth" data-event-label-from="known" novalidate>
```
yarpo: if this is how every form will look like why don't we extract it to macro

```
form(actionName, labelFrom)
```

src/main/public/js/analytics-custom-event-sender.js
```diff
@@ -0,0 +1,69 @@
+$(function () {
+  function sendEvent(eventCategory, eventAction, eventLabel) {
```
yarpo:
```
sendEvent{white space}(eventCategory, eventAction, eventLabel)
```

?

src/main/public/js/analytics-custom-event-sender.js
timja-kainos: actually, its fine, managed to find the data

src/main/features/ccj/views/date-of-birth.njk
geek-soft: For now only selected forms (these with radio buttons) will submit these custom GA events. Not quite sure how it will be in the future. I treat it a bit like experimental feature. When widely adopted we can thing to make it more reusable.

Also using macro is not handy as we would need to do:

```
{{ form(actionName, labelFrom) }}

</form>
```

src/main/public/js/analytics-custom-event-sender.js
geek-soft: Fair point, I just used to call all DOM elements retrieved using jQuery as `xyzElement`. I am happy to change it as it is probably just mine convention.

src/main/public/js/analytics-custom-event-sender.js
geek-soft: Agree, would be better not to throw errors when radio button is not selected.

src/main/features/ccj/views/date-of-birth.njk
yarpo: Aren't you on holidays!? ;)

PR #60, opened 2017-09-15
Make 'must be over 18' validation error message generic

geek-soft:
### JIRA link ###

https://tools.hmcts.net/jira/browse/ROC-2341

### Change description ###

Implementation is based on what I discussed with @BenClancy. If no need to internationalise the application that would be reasonable good solution however with a need for Welsh translation it requires to import `i18next` framework in both `src/main` and `src/test` code.

Also I am not sure the hint is in the right place. There are visually other ways of convey that hint. Would be good to check with @tarik-johnston1 on Monday.

```
[ ] Yes
[x] No
```


geek-soft:
screen shot 2017-09-15 at 22 26 03

screen shot 2017-09-15 at 22 26 31

screen shot 2017-09-15 at 22 30 49

screen shot 2017-09-15 at 22 31 01

PR #61, opened 2017-09-18
ROC-1775 Support additional claimants change in Claim store

dharmendrak:


src/test/a11y/mocks.ts
```diff
@@ -80,7 +80,8 @@ function mockedClaim () {
   const companyDetails = new CompanyDetails()
   companyDetails.address = new Address()
   claim.claimData.defendants = [companyDetails]
-  claim.claimData.claimant = new Individual()
+  let individual = new Individual()
```
timja-kainos: const

dharmendrak: done

PR #67, opened 2017-09-21
Make NSP failure a warning

timja-kainos:
There's currently a nsp issue:
https://nodesecurity.io/advisories/525
With no fix, rather than suppressing it 
and forgetting about it we would prefer to keep getting notifications about it.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -47,7 +47,8 @@ timestamps {
           } catch (ignore) {
             sh "cat nsp-report.txt"
             archiveArtifacts 'nsp-report.txt'
-            error "Node security check failed see the report for the errors"
+            notifyBuildResult channel: channel, color: 'warning',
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -47,7 +47,8 @@ timestamps {
           } catch (ignore) {
             sh "cat nsp-report.txt"
             archiveArtifacts 'nsp-report.txt'
-            error "Node security check failed see the report for the errors"
+            notifyBuildResult channel: channel, color: 'warning',
+              message: 'Node security check failed see the report for the errors'
           }
           sh "rm nsp-report.txt"
         }
```
yarpo: do we want to update yarn.lock?

timja-kainos: I updated it to update our other dependencies, added it as a separate commit

PR #72, opened 2017-09-27
Add more validation on repayment plan

timja-kainos:
This adds support for checking whether the initial instalment and next
instalment is less than or equal to the remaining amount.


src/main/app/forms/validation/validators/isLessThanOrEqualToSumOf.ts
```diff
@@ -0,0 +1,50 @@
...
+      && typeof lessThanAmount === 'number'
+      && typeof amountToSum === 'number'
+      && (value + amountToSum) <= lessThanAmount
+  }
+}
+
+export function IsLessThanOrEqualToSumOf (sum: string, lessThanAmount: string, validationOptions?: ValidationOptions) {
+  return function (object: Object, propertyName: string) {
+    registerDecorator({
+      name: 'isLessThan',
```
geek-soft: I do not think we have to provide a name for the validator. If we do, better to make it in sync with validator name.

src/main/features/ccj/routes/repayment-plan.ts
```diff
@@ -36,6 +36,7 @@ export default express.Router()
         const user: User = res.locals.user
 
         if (form.hasErrors()) {
+          console.log(form.errors)
```
geek-soft: Lets remove it.

src/main/app/forms/validation/validators/isLessThanOrEqualToSumOf.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    return typeof value === 'number'
+      && typeof lessThanAmount === 'number'
+      && typeof amountToSum === 'number'
+      && (value + amountToSum) <= lessThanAmount
+  }
+}
+
+export function IsLessThanOrEqualToSumOf (sum: string, lessThanAmount: string, validationOptions?: ValidationOptions) {
```
geek-soft: I would say it would be nice to have some JSDoc comment saying what this validator is doing / what problem it is supposed to solve since it is quite complex one. Validator name does not fully uncover its complexity.

timja-kainos: I know, couldn't think of a better name :(

src/main/app/forms/validation/validators/isLessThanOrEqualToSumOf.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    const amountToSumProperty = args.constraints[0]
+    const lessThanAmountPropertyName = args.constraints[1]
+
+    const lessThanAmount = (args.object as any)[lessThanAmountPropertyName]
+    const amountToSum = (args.object as any)[amountToSumProperty]
+
+    // Not this validator's concern if the amount hasn't been entered
+    if (!amountToSum) {
```
geek-soft: Is returning valid in case value exceeds `lessThanAmount` but `amountToSum ` has not been specified correct. I would more expect `amountToSum` to fallback to `0` in such scenario which would simply ignore it but validator would still asses that value is less then max limit.

src/test/app/forms/validation/validators/IsLessThanOrEqualToSumOf.ts
```diff
@@ -0,0 +1,51 @@
...
+  return {
+    value: undefined,
+    targetName: undefined,
+    object: { fieldToSum: fieldToSumValue, relatedField: value },
+    property: undefined,
+    constraints: ['fieldToSum', 'relatedField']
+  }
+}
+
+describe('IsLessThanConstraint', () => {
```
geek-soft: Will be easier for use if name of the test block reflects tested class. In that case it would be `IsLessThanOrEqualToSumOfConstraint`.

src/main/app/forms/validation/validators/isLessThanOrEqualToSumOf.ts
timja-kainos: Nope, the fields also have a:
```
@IsPositive({ message: ValidationErrors.INSTALMENTS_AMOUNT_INVALID })
```
Which marks the fields as required basically.
If the above check isn't there then validation gets failed on both fields when it should only fail on the missing field

src/test/app/forms/validation/validators/IsLessThanOrEqualToSumOf.ts
timja-kainos: oops, thought I got everything

src/test/app/forms/validation/validators/IsLessThanOrEqualToSumOf.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+describe('IsLessThanOrEqualToSumOfConstraint', () => {
+  const constraint: IsLessThanOrEqualToSumOfConstraint = new IsLessThanOrEqualToSumOfConstraint()
+
+  describe('validate', () => {
+
+    describe('should return true when ', () => {
+      it('given an undefined value', () => {
+        expect(constraint.validate(undefined)).to.equal(true)
```
geek-soft: That is interesting, I would have though this to throw an `Invalid number of arguments, got...` error.

timja-kainos: the first thing it checks is the value... I can change the order if you want?

src/test/app/forms/validation/validators/IsLessThanOrEqualToSumOf.ts
```diff
@@ -0,0 +1,51 @@
...
+const DEFAULT_FIELD_TO_SUM_VALUE = 30
+const DEFAULT_VALUE = 200
+
+function validationArgs (fieldToSumValue: number, value: number): ValidationArguments {
+  return {
+    value: undefined,
+    targetName: undefined,
+    object: { fieldToSum: fieldToSumValue, relatedField: value },
+    property: undefined,
+    constraints: ['fieldToSum', 'relatedField']
```
geek-soft: That is outdated now, in this context we could probably use empty array `[]`.

Also for the test we probably only need to populate constraint property:

```
function validationArgs (sum: number, lessThenAmount: number): ValidationArguments {
  return {
    constraints: [sum, lessThenAmount]
  } as ValidationArguments
}
```

timja-kainos: that's actually the names of the field in the dummy object  I created for validation

src/test/app/forms/validation/validators/IsLessThanOrEqualToSumOf.ts
geek-soft: I would have expect validator to be properly configured before it starts its job. That would also help us failing fast. I would recommend switching an order. Worth paying an attention to other validators in the future as well.

src/test/app/forms/validation/validators/IsLessThanOrEqualToSumOf.ts
geek-soft: Right, forgot the `object: {}` is actually an instance of class under validation. Ignore me then.

PR #74, opened 2017-09-27
ROC-2371: Store user data in the draft to enable reporting

rlewan:
### JIRA link ###

https://tools.hmcts.net/jira/browse/ROC-2371

### Change description ###

Store user data in the draft to make it available for service manager reports.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/app/models/draft.ts
+++ b/src/main/app/models/draft.ts
@@ -1,3 +1,5 @@
 export class Draft {
   lastUpdateTimestamp: number
+  userEmail: string
+  userName: string
 }
--- a/src/main/common/draft/draftMiddleware.ts
+++ b/src/main/common/draft/draftMiddleware.ts
@@ -1,5 +1,6 @@
 import * as express from 'express'
 import DraftStoreClient from 'common/draft/draftStoreClient'
+import User from 'idam/user'
 
 export class DraftMiddleware<T> {
   client: DraftStoreClient<T>
@@ -19,6 +20,7 @@ export class DraftMiddleware<T> {
           if (!draft) {
             draft = this.deserializeFn(undefined)
           }
+          this.setUserData(draft, res.locals.user)
           res.locals.user[`${this.draftType}Draft`] = draft
           next()
         })
@@ -36,5 +38,10 @@ export class DraftMiddleware<T> {
   delete (res: express.Response, next: express.NextFunction): Promise<void> {
     return this.client
       .delete(res.locals.user.id)
+  }
+
+  private setUserData (draft: any, user: User) {
+    draft.userEmail = user.email
+    draft.userName = `${user.forename} ${user.surname}`
   }
 }
--- a/src/main/common/draft/draftMiddleware.ts
+++ b/src/main/common/draft/draftMiddleware.ts
@@ -1,8 +1,9 @@
 import * as express from 'express'
 import DraftStoreClient from 'common/draft/draftStoreClient'
 import User from 'idam/user'
+import { Draft } from 'models/draft'
 
-export class DraftMiddleware<T> {
+export class DraftMiddleware<T extends Draft> {
   client: DraftStoreClient<T>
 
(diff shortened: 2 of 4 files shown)


src/main/common/draft/draftMiddleware.ts
```diff
@@ -11,18 +13,19 @@ export class DraftMiddleware<T> {
...
-        .then(draft => {
-          if (!draft) {
-            draft = this.deserializeFn(undefined)
-          }
-          res.locals.user[`${this.draftType}Draft`] = draft
-          next()
-        })
-        .catch(next)
+      try {
+        let draft: T = await this.client.retrieve(res.locals.user.id, this.deserializeFn)
```
timja-kainos: Should this be `Draft`?

src/main/app/models/draft.ts
```diff
@@ -1,3 +1,5 @@
 export class Draft {
   lastUpdateTimestamp: number
+  userEmail: string
+  userName: string
```
timja-kainos: little bit of a misnomer.
`idamName`?
`submitterName`?

geek-soft: I would prefer to have it wrapped with some `owner` property. Then we have:

```
{
  ...
 owner: {
  email: "",
  name: ""
 }
}
```

or even more separate by having:

```
{
  ...
 metadata: {
  owner: {
   email: "",
   name: ""
  }
 }
}
```

By the way `lastUpdateTimestamp` will get removed from document when we switch to V3.

src/main/common/draft/draftMiddleware.ts
```diff
@@ -11,18 +13,19 @@ export class DraftMiddleware<T> {
...
+        if (!draft) {
+          draft = this.deserializeFn(undefined)
+        }
+        this.setUserData(draft, res.locals.user)
```
geek-soft: As I understand that will always override persistence user data when case is retrieved, which gets potentially persisted next time save is made. That is a bit wired / unexpected behaviour for me. 

I would have liked to initiate draft metadata when draft is created for the first time and do not modify it afterwards.

timja-kainos: hmm, if we only initialise then we wouldn't get the data for old draft's when the user logs in

geek-soft:
Maybe in long term we could have some java reporting solution where we can nicely fetch data from other sources like IDAM. I know this is still possible with bash scripts, just not as nice to maintain when complexity arise.

Long term also IDAM should allow use to get user details by user ID so we would need to story only owner identifier. IDAM was happy for us to make the change to create such `getUserDetailsByID` endpoint.

src/main/common/draft/draftMiddleware.ts
geek-soft: I saw this issue but proposed solution looks like a big sacrifice to overcome the problem. Maybe we do not have to care about few existing drafts. Or maybe such metadata should be only saved at persist time and read as is, if does not exist then fine UI does not need it.

For me cleaner is to call `this.setUserData(draft, res.locals.user)` at save time so then owner is overridden with each save which makes it really a last owner / modifier.

geek-soft: A `DraftStoreClient.save` would be the right place, the same as we set `draft.lastUpdateTimestamp`

timja-kainos:
when were they happy for us to make that change?
last I heard they weren't
and nilesh wasn't keen on fetching data from idam for reporting because of the number of api calls we would need to make

geek-soft:
Regarding IDAM change that was around the time we had a need to retrieve user details by ID. The condition at that time was not to return whole user record but just basic properties. Unfortunately we did not purse this at that time and possible there was change of heart due to FRIDAM related reasons.

Volume concern seems valid however we would probably find a solution for it. That would guarantee the reaching for the most up to date details however I am not agains having basic user properties as a part of draft as long as for now.

rlewan:
Closing this as Draft Store-based reports are halted for now.

PR #75, opened 2017-09-28
Don't allow names to be edited

timja-kainos:
Fix bugs around loading of contactPerson and dateOfBirth


src/main/features/response/views/your-details.njk
```diff
@@ -16,30 +16,30 @@
...
             ]
           %}
           {% set addressSectionHeading = 'Your address' %}
         {% elif form.model.type == 'soleTrader' %}
           {%
             set fields = [
-              { label: 'Your full name', name: 'name' },
-              { label: 'Business name (optional)', name: 'businessName' }
+              { readOnly: true, label: 'Your full name', value: claim.claimData.defendant.name },
+              { readOnly: true, label: 'Business name (optional)', value: claim.claimData.defendant.businessName }
```
geek-soft: I wonder if this still make sense to say `(optional)` for read only fields. For me that is the hint for someone who is filling the form. If value not provided then that heading won't be even displayed.

src/main/app/forms/models/companyDetails.ts
```diff
@@ -12,7 +12,7 @@ export class CompanyDetails extends PartyDetails {
 
   @IsDefined({ message: ValidationErrors.CONTACT_PERSON_REQUIRED, groups: ['claimant'] })
   @IsNotBlank({ message: ValidationErrors.CONTACT_PERSON_REQUIRED, groups: ['claimant'] })
-  @MaxLength(35, { message: ValidationErrors.CONTACT_PERSON_NAME_TOO_LONG, groups: ['claimant'] })
+  @MaxLength(35, { message: ValidationErrors.CONTACT_PERSON_NAME_TOO_LONG, groups: ['claimant', 'response'] })
```
geek-soft: That makes sense, something we forgot earlier 👍

src/main/features/response/routes/your-details.ts
```diff
@@ -44,7 +47,18 @@ export default express.Router()
 
     const partyDetails: PartyDetails = plainToClass(PartyDetails, user.claim.claimData.defendant)
     if (user.responseDraft.defendantDetails.partyDetails) {
-      partyDetails.name = user.responseDraft.defendantDetails.partyDetails.name
+      switch (user.responseDraft.defendantDetails.partyDetails.type) {
```
geek-soft: Looks like fix for something we missed the other day.

I wonder if we could make it even better by pointing the right type to `plainToClass(<here>, user.claim.claimData.defendant)`.

Then we could extract type from `user.claim.claimData.defendant.type`to find constructor for given party type.

To summarise something like `plainToClass(PartyDetailsFactory.constructorFor(user.claim.claimData.defendant.type), user.claim.claimData.defendant)`

src/main/features/response/views/your-details.njk
timja-kainos: will doublecheck

timja-kainos: confirmed, remove optional with Art

src/main/features/response/routes/your-details.ts
timja-kainos: will give it a go, but do need to get this in soon for release next week +1

geek-soft: Theoretically it should work. I thought that when pass any constructor for class extending from `TheirDetails` it should nicely deserialise it but I never tested subclasses with this library. Happy to leave it as is for now and rework it when we start using the library more extensively.

timja-kainos: subclassing does work

PR #84, opened 2017-10-02
ROC-2400 Allow having multiple response / CCJ drafts

geek-soft:
### JIRA link ###

https://tools.hmcts.net/jira/browse/ROC-2400

### Change description ###

This PR is to allow us having multiple drafts for response or CCJ. To make it possible I decided to switch to new draft-store API. Unfortunately new API comes with extract authentication requirements (user + service auth) so there is a few extra changes to cover it.

TODO:
- conflicts resolve
- ensure backward compatibility
- fix / add tests

```
[x] Yes - new version of the draft-store has to be deployed
[ ] No
```


src/main/common/security/serviceTokenFactory.ts
```diff
@@ -0,0 +1,15 @@
+import IdamClient from 'idam/idamClient'
+
+import ServiceAuthToken from 'idam/serviceAuthToken'
+
+let token: ServiceAuthToken
```
timja-kainos: is this just cached for the request? or longer?

geek-soft: Per app since when `serviceTokenFactory.ts` is imported Node.js caches imports and JS file is executed just once.

src/main/app/drafts/models/draftClaim.ts
```diff
@@ -6,10 +6,10 @@ import InterestDate from 'forms/models/interestDate'
 import Reason from 'forms/models/reason'
 import * as uuid from 'uuid'
 import { Defendant } from 'app/drafts/models/defendant'
-import { Draft } from 'app/models/draft'
+import { DraftDocument } from 'app/models/draft'
 import { QualifiedStatementOfTruth } from 'app/forms/models/qualifiedStatementOfTruth'
 
-export default class DraftClaim extends Draft implements Serializable<DraftClaim> {
+export default class DraftClaim extends DraftDocument implements Serializable<DraftClaim> {
```
yarpo: maybe we should also rename this file?

src/main/app/idam/serviceAuthToken.ts
```diff
@@ -1,5 +1,13 @@
...
+import { JwtUtils } from 'common/utils/jwtUtils'
+
 export default class ServiceAuthToken {
   constructor (public bearerToken: string) {
     this.bearerToken = bearerToken
   }
+
+  hasExpired (): boolean {
+    const { exp } = JwtUtils.decodePayload(this.bearerToken)
+    return moment().unix() > exp
```
yarpo: when `exp` === moment().unix() is it not expired?

src/main/common/draft/draftMiddleware.ts
```diff
@@ -1,40 +1,77 @@
...
-  constructor (public draftType: string, public deserializeFn: (value: any) => T = (value) => value) {
-    if (!draftType || draftType.trim() === '') {
-      throw new Error('Draft type is required to instantiate middleware')
-    }
-    this.client = new DraftStoreClient<T>(draftType)
+  const externalIds: RegExpMatchArray = path.match(pattern)
+  if (externalIds !== null) {
+    return externalIds[1]
   }
+  return undefined
```
yarpo: not exception?

src/main/common/draft/draftStoreClient.ts
```diff
@@ -1,55 +1,67 @@
...
+    return request
+      .get(endpointURL, {
+        qs: qs,
+        headers: this.authHeaders(userAuthToken)
+      })
+      .then((response: any) => {
+        return response.data
+          .filter(draft => type ? draft.type === type : true)
+          .map(draft => {
+            const wrapper = new Draft()
```
yarpo: I'm not sure if intelij can understand that `wrapper` is type of `Draft`?

maybe:

```
const wrapper:Draft = new Draft()
```

src/main/common/draft/draftStoreClient.ts
```diff
@@ -1,55 +1,67 @@
...
-      .then(draft => deserializationFn(draft))
-      .catch(err => {
-        if (err.statusCode === HttpStatus.NOT_FOUND) {
-          return undefined
-        } else {
-          throw err
-        }
-      })
+  delete (draft: Draft<T>, userAuthToken: string): Promise<void> {
+    return request.delete(`${endpointURL}/${draft.id}`, {
```
yarpo: shouldn't we make sure `id` is defined number?

src/test/a11y/mocks.ts
```diff
@@ -178,11 +170,27 @@ mock('fees/feesClient', {
...
-      res.locals.user.responseDraft = mockedResponseDraft()
-      next()
+mock('common/draft/draftMiddleware', {
+  'DraftMiddleware': {
+    requestHandler: (draftType: string) => {
+      return (req: express.Request, res: express.Response, next: express.NextFunction): void => {
+        const draft = new Draft<DraftDocument>()
+        switch (draftType) {
+          case 'claim':
+            draft.document = mockedDraftClaim()
```
yarpo: maybe not for this PR, but `mockedDraftClaim`, `mockedResponseDraft` and `mockCCJDraft` don't look consistent. [position of "Draft"]

src/main/common/draft/draftStoreClient.ts
timja-kainos: it's pretty obvious ;) and intellij can understand it

src/main/app/drafts/models/draftClaim.ts
timja-kainos: why? the class name hasn't changed

src/main/app/models/draft.ts
```diff
@@ -1,3 +1,13 @@
-export class Draft {
-  lastUpdateTimestamp: number
+import { Moment } from 'moment'
+
+export class DraftDocument {
```
timja-kainos: I would consider moving this to its own file

src/main/common/draft/draftMiddleware.ts
```diff
@@ -1,40 +1,77 @@
...
+import { Draft, DraftDocument } from 'models/draft'
+import { DraftStoreClientFactory } from 'common/draft/draftStoreClientFactory'
 
-export class DraftMiddleware<T> {
-  client: DraftStoreClient<T>
+/**
+ * Extracts external ID from paths in the '/case/<external-id>.*' format
+ */
+function extractExternalId (path: string): string {
+  const pattern = /\/case\/([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})\/.*/
```
timja-kainos: This is already done in the claimMiddleware, this should be extracted somewhere common and then reused in both places

src/main/common/draft/draftStoreClient.ts
timja-kainos: Could these fields be initialised in the constructor?

src/main/app/idam/serviceAuthToken.ts
geek-soft: Good point, JWT spec says:

> The "exp" (expiration time) claim identifies the expiration time on
   or after which the JWT MUST NOT be accepted for processing.  The
   processing of the "exp" claim requires that the current date/time
   MUST be before the expiration date/time listed in the "exp" claim.

src/main/common/draft/draftMiddleware.ts
geek-soft: I left this undefined as this method might not return UUID in case URL does not have it. Will add this detail to the JSDoc

src/main/common/draft/draftStoreClient.ts
geek-soft: I decided not to give a time since you can see it at the same line. As a interesting thing a TSLint used by Angular is configured in such a way that it will return errors if you defining type in such an obvious cases.

src/main/common/draft/draftStoreClient.ts
geek-soft: Good spot, will add this as a optimisation to prevent API calls when non saved draft is being deleted.

src/test/a11y/mocks.ts
geek-soft: Good idea, done.

src/main/common/draft/draftMiddleware.ts
geek-soft: Missed that one, good idea to reuse. Created `UUIDUtils.extractFrom` for that purpose :)

src/main/common/draft/draftStoreClient.ts
geek-soft: Good idea, created constructor 👍

PR #89, opened 2017-10-04
ROC-2069 Defendant change defence response option

rattur:
Defendant response page options and flow change
 
### JIRA link ###

https://tools.hmcts.net/jira/browse/ROC-2069

### Change description ###

```
[ ] Yes
[x ] No
```


timja-kainos:
conflicts

src/main/features/response/views/reject-all-of-claim.njk
```diff
@@ -0,0 +1,55 @@
...
+{% set heading = t("I reject all of the claim") %}
+
+{% block content %}
+  <div class="grid-row">
+    <div class="column-two-thirds">
+
+      <form method="post" novalidate>
+        {{ csrfProtection(csrf) }}
+
+        {% set error = form.errorFor('type') %}
```
timja-kainos: why is this here

src/main/features/response/views/reject-all-of-claim.njk
```diff
@@ -0,0 +1,55 @@
...
+        <div class="form-group {% if error %} form-group-error {% endif %}">
+          <fieldset id="type[label]">
```
timja-kainos: pretty sure radiogroup comes with a fieldset

src/main/features/response/views/reject-part-of-claim.njk
```diff
@@ -0,0 +1,48 @@
...
+
+{% set heading = t("I reject part of the claim") %}
+
+{% block content %}
+  <div class="grid-row">
+    <div class="column-two-thirds">
+
+      <form method="post" novalidate>
+        {{ csrfProtection(csrf) }}
+        {% set error = form.errorFor('type') %}
```
timja-kainos: this shouldn't be here

src/main/features/response/views/reject-part-of-claim.njk
```diff
@@ -0,0 +1,48 @@
...
+        <div class="form-group {% if error %} form-group-error {% endif %}">
+          <fieldset id="type[label]">
+            <legend class="visually-hidden">{{ t("I reject part of the claim") }}</legend>
+            {% if error %}<span class="error-message">{{ t(error) }}</span>{% endif %}
```
timja-kainos: down to here from above comment shouldn't be here

src/main/features/response/routes/response-type.ts
```diff
@@ -36,9 +36,9 @@ export default express.Router()
         await ResponseDraftMiddleware.save(res, next)
 
         if (ResponseType.OWE_NONE === form.model.type) {
-          res.redirect(Paths.defenceOptionsPage.evaluateUri({ externalId: externalId }))
```
timja-kainos: this check isn't right, there's 3 pages it could go to:
part admission
full rejection
task-list

I just pulled your branch and the full admission button is going to partial admission

src/main/features/response/routes/check-and-send.ts
```diff
@@ -103,19 +102,16 @@ export default express.Router()
         switch (responseType) {
           case ResponseType.OWE_NONE:
             if (defendantIsCounterClaiming(user)) {
-              res.redirect(Paths.counterClaimPage.evaluateUri({ externalId: user.claim.externalId }))
+              res.redirect(Paths.defenceRejectAllOfClaimPage.evaluateUri({ externalId: user.claim.externalId }))
```
timja-kainos: this isn't right its going to a page with radio buttons, it should still go to the guidance page until we can handle it.

src/main/features/response/routes/check-and-send.ts
```diff
@@ -103,19 +102,16 @@ export default express.Router()
...
               return
             }
             break
           case ResponseType.OWE_SOME_PAID_NONE:
-          case ResponseType.OWE_ALL_PAID_SOME:
-            res.redirect(Paths.partialAdmissionPage.evaluateUri({ externalId: user.claim.externalId }))
+            res.redirect(Paths.defenceRejectPartOfClaimPage.evaluateUri({ externalId: user.claim.externalId }))
```
timja-kainos: this isn't right its going to a page with radio buttons, it should still go to the guidance page until we can handle it.

src/main/features/response/routes/check-and-send.ts
```diff
@@ -103,19 +102,16 @@ export default express.Router()
...
             return
           case ResponseType.OWE_ALL_PAID_NONE:
-            res.redirect(Paths.fullAdmissionPage.evaluateUri({ externalId: user.claim.externalId }))
+            res.redirect(Paths.defenceRejectPartOfClaimPage.evaluateUri({ externalId: user.claim.externalId }))
```
timja-kainos: this isn't right its going to a page with radio buttons, it should still go to the guidance page until we can handle it.

src/main/features/response/views/response-type.njk
```diff
@@ -16,30 +16,23 @@
           <legend class="visually-hidden">{{ t("Do you owe the money claimed?") }}</legend>
           {% if error %}<span class="error-message">{{ t(error) }}</span>{% endif %}
 
-          {{ radioOption(form = form, name = 'type[value]', value = 'OWE_ALL_PAID_ALL',
-            label = t('I’ve already paid'), bold = true,
-            hint = t("You need to submit a defence explaining how much you’ve paid the claimant and when you paid.")
-          )}}
+          <p>{{ t("You can make a counterclaim against the claimant after you submit your response.") }}</p>
```
timja-kainos: you are missing the needed configuration for the analytics sender to send the data to google analytics
see this PR:
https://github.com/hmcts/cmc-citizen-frontend/pull/58/files

src/main/features/response/views/reject-all-of-claim.njk
```diff
@@ -0,0 +1,55 @@
...
+        },
+        {
+          label: "I dispute the claim",
+          value: "dispute",
+          hint: "",
+          bold: true
+        },
+        {
+          label: 'Defend now and also make a counterclaim',
+          value: "counterClaim",
```
geek-soft: I do not see this being used anywhere in the application. How does is fit into the counter claim feature we already have? Either start using it or delete.

src/main/features/response/routes/reject-part-of-claim.ts
```diff
@@ -0,0 +1,39 @@
...
+import User from 'idam/user'
+
+function renderView (form: Form<RejectPartOfClaim>, res: express.Response) {
+  res.render(Paths.defenceRejectPartOfClaimPage.associatedView, {
+    form: form
+  })
+}
+
+export default express.Router()
+  .get(Paths.defenceRejectPartOfClaimPage.uri, ErrorHandling.apply(async (req: express.Request, res: express.Response, next: express.NextFunction) => {
```
geek-soft: That route has to be protected by guard so it is impossible to access it unless somebody selected `OWE_SOME_PAID_NONE` on the response type page.

Remember to cover both GET and POST handlers and reflect that fact in tests as well.

src/main/features/response/routes/reject-all-of-claim.ts
```diff
@@ -0,0 +1,46 @@
...
+import User from 'idam/user'
+
+function renderView (form: Form<RejectAllOfClaim>, res: express.Response) {
+  res.render(Paths.defenceRejectAllOfClaimPage.associatedView, {
+    form: form
+  })
+}
+
+export default express.Router()
+  .get(Paths.defenceRejectAllOfClaimPage.uri, ErrorHandling.apply(async (req: express.Request, res: express.Response, next: express.NextFunction) => {
```
geek-soft: That route has to be protected by guard so it is impossible to access it unless somebody selected `OWE_NONE` on the response type page.

Remember to protected both GET and POST handlers and reflect that fact in tests as well.

src/main/features/response/views/reject-all-of-claim.njk
```diff
@@ -0,0 +1,55 @@
...
+        {% set error = form.errorFor('type') %}
+        <div class="form-group {% if error %} form-group-error {% endif %}">
+          <fieldset id="type[label]">
+            <legend class="visually-hidden">{{ t("I reject all of the claim") }}</legend>
+            {% if error %}<span class="error-message">{{ t(error) }}</span>{% endif %}
+        {{
+        radioGroup(
+        label = "Why do you reject all the claim?",
+        name = "option",
+        options = [
```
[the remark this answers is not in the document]
rattur: Yes I will address this, need to change statement of truth template too

geek-soft:
PR needs merge with master. This RP is build on out-of-sync master which does not even have changed structure of the drafts.

src/main/features/response/routes/response-type.ts
```diff
@@ -36,9 +35,13 @@ export default express.Router()
         await DraftService.save(user.responseDraft, user.bearerToken)
 
         if (ResponseType.OWE_NONE === form.model.type) {
```
geek-soft: Using `switch` here makes more sense, it should also make code a bit nicer / easier to read.

src/main/features/response/draft/responseDraft.ts
```diff
@@ -32,6 +36,8 @@ export class ResponseDraft extends DraftDocument implements Serializable<Respons
       if (input.qualifiedStatementOfTruth) {
         this.qualifiedStatementOfTruth = new QualifiedStatementOfTruth().deserialize(input.qualifiedStatementOfTruth)
       }
+      this.rejectPartOfClaim = new RejectPartOfClaim(input.rejectPartOfClaim && input.rejectPartOfClaim.option)
+      this.rejectAllOfClaim = new RejectPartOfClaim(input.rejectAllOfClaim && input.rejectAllOfClaim.option)
```
geek-soft: Type does not match here, you should have used `RejectAllOfClaim` here.

src/main/features/response/guards/oweSomePaidNoneResponseRequiredGuard.ts
```diff
@@ -0,0 +1,19 @@
+import * as express from 'express'
+
+import { ResponseType } from 'response/form/models/responseType'
+import { Paths } from 'response/paths'
+import User from 'idam/user'
+
+export default class OweSomePaidNoneResponseRequiredGuard {
```
geek-soft: Please do not use default import. Using default imports make refactoring / code maintenance slightly more difficult. Please change to named import.

geek-soft: That class seems not to be used at all. What is the point of having guard / a class which is not used?

src/main/features/response/tasks/oweMoneyTask.ts
```diff
@@ -7,6 +7,13 @@ export class OweMoneyTask {
       return false
     }
 
+    if (responseDraft.response.type !== ResponseType.OWE_ALL_PAID_NONE) {
```
geek-soft: That does not even change anything. I can still choose an option on response type page, be taken to the response details screen, jump to task list without doing anything and see COMPLETED. That is not CORRECT. Task is completed only when all necessary / mandatory details has been captured.

Please correct the logic and reflect it in the unit test.

src/main/features/response/views/reject-all-of-claim.njk
```diff
@@ -0,0 +1,51 @@
...
+{% set heading = t("I reject all of the claim") %}
+
+{% block content %}
+  <div class="grid-row">
+    <div class="column-two-thirds">
+
+      <form method="post" novalidate>
+        {{ csrfProtection(csrf) }}
+
+        <div class="form-group {% if error %} form-group-error {% endif %}">
```
geek-soft: That is not needed as `radioGroup` macro does it for us. Please check what `radioGroup` is doing and remove unnecessary markup.

src/main/features/response/views/reject-all-of-claim.njk
```diff
@@ -0,0 +1,51 @@
...
+
+            <legend class="visually-hidden">{{ t("I reject all of the claim") }}</legend>
+            {% if error %}<span class="error-message">{{ t(error) }}</span>{% endif %}
```
geek-soft: This is not needed as well.

src/main/features/response/views/reject-all-of-claim.njk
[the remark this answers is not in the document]
geek-soft: You did say that you will address it however it is still not addressed.

src/main/features/response/views/reject-all-of-claim.njk
```diff
@@ -0,0 +1,51 @@
...
+        {
+          label: 'Defend now and also make a counterclaim',
+          value: "counterClaim",
+          hint: 'You’ll need to complete the defence and counterclaim form.'
+        }
+        ],
+        form = form
+        )
+        }}
+          <div class="form-group">
```
geek-soft: We do not put buttons into `form-group` div. Any reason you did it? If not please remove.

src/main/features/response/views/reject-part-of-claim.njk
```diff
@@ -0,0 +1,46 @@
...
+
+{% set heading = t("I reject part of the claim") %}
+
+{% block content %}
+  <div class="grid-row">
+    <div class="column-two-thirds">
+
+      <form method="post" novalidate>
+        {{ csrfProtection(csrf) }}
+        <div class="form-group {% if error %} form-group-error {% endif %}">
```
geek-soft: The same here and other places in the file.

src/main/views/includes/scripts.njk
```diff
@@ -6,7 +6,7 @@
 <script src="{{ asset_paths['js'] }}/application.js"></script>
 <script src="{{ asset_paths['js'] }}/detectJavascript.js"></script>
 <script src="{{ asset_paths['js'] }}/analytics.js"></script>
-<script src="{{ asset_paths['js'] }}/analytics-custom-event-sender.js"></script>
+<script src="{{ asset_paths['js'] }}/analytics-custom-event-sender.js?2017-10-11"></script>
```
geek-soft: Should not be needed but preferred to add it in case. In the future asset fingerprinting should take care of it.

src/test/a11y/mocks.ts
```diff
@@ -63,3 +63,11 @@ mock('ccj/guards/ccjGuard', {
 mock('ccj/guards/individualDateOfBirthGuard', {
   'IndividualDateOfBirthGuard': justForwardRequestHandler
 })
+
+mock('response/guards/guardFactory', {
+  GuardFactory: {
```
geek-soft: In the future this generic guard factory should help us keeping route specific guards close to the routes and extract guards to separate files only when these need to be shared across many routes

src/main/features/response/form/models/rejectAllOfClaim.ts
```diff
@@ -0,0 +1,36 @@
...
+      RejectAllOfClaimOption.DISPUTE,
+      RejectAllOfClaimOption.COUNTER_CLAIM
+    ]
+  }
+
+  static except (option: string): string[] {
+    if (option === undefined) {
+      throw new Error('Option is required')
+    }
+    return RejectAllOfClaimOption.all().filter(_ => _ !== option)
```
rlewan: Can we use a word as an argument name instead of `_`?

src/main/features/response/routes/reject-all-of-claim.ts
```diff
@@ -0,0 +1,53 @@
...
+export default express.Router()
+  .get(
+    Paths.defenceRejectAllOfClaimPage.uri,
+    GuardFactory.create(isRequestAllowed, accessDeniedCallback),
+    ErrorHandling.apply(async (req: express.Request, res: express.Response, next: express.NextFunction) => {
+      renderView(new Form(res.locals.user.responseDraft.document.rejectAllOfClaim), res)
+    }))
+  .post(
+    Paths.defenceRejectAllOfClaimPage.uri,
+    GuardFactory.create(isRequestAllowed, accessDeniedCallback),
```
rlewan: This guard can be created once as a constant and then passed in as a parameter.

src/main/features/response/routes/reject-part-of-claim.ts
```diff
@@ -0,0 +1,53 @@
...
+export default express.Router()
+  .get(
+    Paths.defenceRejectPartOfClaimPage.uri,
+    GuardFactory.create(isRequestAllowed, accessDeniedCallback),
+    ErrorHandling.apply(async (req: express.Request, res: express.Response, next: express.NextFunction) => {
+      renderView(new Form(res.locals.user.responseDraft.document.rejectPartOfClaim), res)
+    }))
+  .post(
+    Paths.defenceRejectPartOfClaimPage.uri,
+    GuardFactory.create(isRequestAllowed, accessDeniedCallback),
```
rlewan: Same here, the guard can be created once.

src/main/features/response/routes/response-type.ts
```diff
@@ -35,10 +35,20 @@ export default express.Router()
...
+            res.redirect(Paths.defenceRejectAllOfClaimPage.evaluateUri({ externalId: externalId }))
+            break
+          case ResponseType.OWE_SOME_PAID_NONE:
+            res.redirect(Paths.defenceRejectPartOfClaimPage.evaluateUri({ externalId: user.claim.externalId }))
+            break
+          case ResponseType.OWE_ALL_PAID_NONE:
+            res.redirect(Paths.taskListPage.evaluateUri({ externalId: user.claim.externalId }))
+            break
+          default:
+            next(new Error('Unknown response type: ' + responseType))
```
rlewan: Backtics can be used here to interpolate the string.

src/main/features/response/routes/reject-part-of-claim.ts
```diff
@@ -0,0 +1,53 @@
...
+
+import { ResponseType } from 'response/form/models/responseType'
+import { RejectPartOfClaim } from 'response/form/models/rejectPartOfClaim'
+import { DraftService } from 'common/draft/draftService'
+import { ErrorHandling } from 'common/errorHandling'
+import User from 'idam/user'
+import { GuardFactory } from 'response/guards/guardFactory'
+
+function isRequestAllowed (res: express.Response): boolean {
+  return res.locals.user.responseDraft.document.response !== undefined
```
yarpo: why don't we assign:

```
const document: responseDraft = res.locals.user.responseDraft.document
```

?

src/main/features/response/views/reject-all-of-claim.njk
```diff
@@ -0,0 +1,46 @@
...
+        <legend class="visually-hidden">{{ t('I reject all of the claim') }}</legend>
+
+        {{
+          radioGroup(
+          label = 'Why do you reject all the claim?',
+          name = 'option',
+          options = [
+            {
+              label: 'I’ve already paid the claim in full',
+              value: 'alreadyPaid'
```
rlewan: Seeing this (not only here but on all such pages as well) I have a feeling we should export all constants classes as nunjucks locals and use them instead of using magic values.

src/main/features/response/views/check-and-send.njk
```diff
@@ -39,18 +39,16 @@
...
-      {% endif %}
-
-      {{ submitButton('Submit response') }}
-      <input type="hidden" name="type" value="{{ form.model.type }}">
-    </div>
+    {{
+      statementOfTruth(
+        form,
+        label = 'I believe that the facts stated in this claim are true',
+        isQualified = (signatureType == 'qualified')
```
yarpo: `===` ?

src/main/public/js/analytics-custom-event-sender.js
```diff
@@ -13,6 +13,15 @@ $(function () {
     ga('send', 'event', eventCategory, eventAction, eventLabel)
   }
 
+  function escape (value) {
+    if (!value) {
```
rlewan: We should also check that the value is a string.

src/test/features/response/draft/responseDraft.ts
```diff
@@ -47,33 +47,53 @@ describe('ResponseDraft', () => {
...
+  })
 
-        expect(responseDraftModel.requireDefence()).to.be.eq(false)
-      })
-      it('should return false when response type is OWE_NONE and counter claim is true', () => {
-        const responseDraftModel: ResponseDraft = new ResponseDraft()
-        responseDraftModel.response = new Response(ResponseType.OWE_NONE)
-        responseDraftModel.counterClaim = new CounterClaim(true)
+  describe('requireDefence', () => {
+    it('should return false when no response type set', () => {
```
rlewan: Either this method is named poorly or there is a bug in its logic. I would expect a method named `requireDefence` to return true if the response is not set.

src/main/features/response/views/reject-all-of-claim.njk
geek-soft: Agree, asked @rattur for the same but it got lower priority.

src/main/features/response/form/models/rejectAllOfClaim.ts
geek-soft: Done

src/main/features/response/routes/reject-all-of-claim.ts
geek-soft: Done

src/main/features/response/routes/reject-part-of-claim.ts
geek-soft: Done

src/main/features/response/routes/response-type.ts
geek-soft: Done

src/main/features/response/views/check-and-send.njk
geek-soft: Done

src/main/features/response/routes/reject-part-of-claim.ts
geek-soft: We could, but I would like to create type definition more so that `res.locals.user.responseDraft.document` gives us right type. Having this we would not need to reassign to variable so often.

src/main/public/js/analytics-custom-event-sender.js
geek-soft: Done.

PR #90, opened 2017-10-04
ROC-2440 Check before you submit

timja-kainos:


src/main/features/claim/views/incomplete-submission.njk
```diff
@@ -0,0 +1,21 @@
+{% extends "layout.njk" %}
+
+{% set title = t('Incomplete submission') %}
+
+{% block content %}
+  <div class="grid-row">
+    <div class="column-two-thirds">
+      <h1 class="heading-large">{{ t('You need to complete all sections before you submit your claim') }}</h1>
```
geek-soft: Lines 8-17 could go into `cmc-commons-frontend` as compost element. The fragment could look something like this:

```
{{ remainingTasksFragment(heading, tasks, taskListLink) }}
````

I would like us share not only low level macros like form controls but also more complex  page fragments.

File: src/main/features/claim/views/incomplete-submission.njk [hunk 2]
```diff
@@ -0,0 +1,21 @@
...
+      <p>{{ t('You still need to complete:') }}</p>
+
+      <ul class="list list-bullet">
+        {% for item in tasks %}
+          <li>{{ item.name }}</li>
```
geek-soft: Task name should be translated.

src/main/features/claim/views/incomplete-submission.njk
```diff
@@ -0,0 +1,21 @@
...
+        {% endfor %}
+      </ul>
+
+      <a href="{{ taskListUri }}">Back to task list</a>
```
geek-soft: Back link should be translated as well.

src/main/features/claim/views/incomplete-submission.njk
timja-kainos: done

src/main/features/claim/views/incomplete-submission.njk
timja-kainos: done

src/main/features/claim/views/incomplete-submission.njk
timja-kainos: done

src/main/features/claim/helpers/taskListBuilder.ts
```diff
@@ -0,0 +1,42 @@
...
+import TaskListItem from 'drafts/tasks/taskListItem'
+import { Paths } from 'claim/paths'
+import { ResolveDispute } from 'drafts/tasks/resolveDispute'
+import { CompletingYourClaim } from 'drafts/tasks/completingYourClaim'
+import { ClaimAmount } from 'drafts/tasks/claimAmount'
+import { ClaimDetails } from 'drafts/tasks/claimDetails'
+import { YourDetails } from 'app/drafts/tasks/yourDetails'
+import { TheirDetails } from 'app/drafts/tasks/theirDetails'
+
+export class TaskListBuilder {
```
geek-soft: That is good refactoring. Something I was thinking for some time however in my head I took it even further. I would like to have factory method which returns fully configured `TaskList` (not just list of `TaskListItem`). That would be used in the first place to render the task list (using common task list fragment shared across all parts of the app which iterates through model and renders both task list headlines as well as tasks itself). That would be used as well in guards and other places where task list has to be questioned. 

I understand that refactoring is beyond what user story defines so I am pointing this as guideline which you can follow if you wanna refactor it even more but cannot ask you to do within the story.

PR #96, opened 2017-10-05
ROC-2061: Add guard forbidding access to response journey when CCJ was requested

yarpo:
### JIRA link ###
https://tools.hmcts.net/jira/browse/ROC-2061

### Change description ###
Added guard for response pages triggered when CCJ was requested

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/features/response/guards/countyCourtJudgmentRequestedGuard.ts
+++ b/src/main/features/response/guards/countyCourtJudgmentRequestedGuard.ts
@@ -1,0 +1,17 @@
+import * as express from 'express'
+import { Paths as DashboardPaths } from 'dashboard/paths'
+import User from 'idam/user'
+const logger = require('@hmcts/nodejs-logging').getLogger('response/guards/countyCourtJudgmentRequestedGuard')
+
+export default class CountyCourtJudgmentRequestedGuard {
+
+  static requestHandler (req: express.Request, res: express.Response, next: express.NextFunction) {
+    const user: User = res.locals.user
+    if (user.claim.countyCourtJudgmentRequestedAt) {
+      logger.warn('State guard: CCJ already requested - redirecting to dashboard')
+      return res.redirect(DashboardPaths.dashboardPage.uri)
+    } else {
+      next()
+    }
+  }
+}
--- a/src/main/features/response/guards/countyCourtJudgmentRequestedGuard.ts
+++ b/src/main/features/response/guards/countyCourtJudgmentRequestedGuard.ts
@@ -3,7 +3,7 @@ import { Paths as DashboardPaths } from 'dashboard/paths'
 import User from 'idam/user'
 const logger = require('@hmcts/nodejs-logging').getLogger('response/guards/countyCourtJudgmentRequestedGuard')
 
-export default class CountyCourtJudgmentRequestedGuard {
+export class CountyCourtJudgmentRequestedGuard {
 
   static requestHandler (req: express.Request, res: express.Response, next: express.NextFunction) {
     const user: User = res.locals.user
--- a/src/main/features/response/index.ts
+++ b/src/main/features/response/index.ts
@@ -10,7 +10,7 @@ import { AlreadyRespondedGuard } from 'response/guards/alreadyRespondedGuard'
 import { ClaimMiddleware } from 'app/claims/claimMiddleware'
 import { DraftMiddleware } from 'common/draft/draftMiddleware'
 import { ResponseDraft } from 'response/draft/responseDraft'
-import CountyCourtJudgmentRequestedGuard from 'response/guards/countyCourtJudgmentRequestedGuard'
+import { CountyCourtJudgmentRequestedGuard } from 'response/guards/countyCourtJudgmentRequestedGuard'
 
 function defendantResponseRequestHandler (): express.RequestHandler {
   function accessDeniedCallback (req: express.Request, res: express.Response): void {
--- a/src/main/features/response/index.ts
+++ b/src/main/features/response/index.ts
@@ -34,10 +34,7 @@ export class Feature {
       /^\/case\/.+\/response\/(?![\d]+\/receiver|confirmation|full-admission|partial-admission|counter-claim|receipt).*$/,
       AlreadyRespondedGuard.requestHandler
     )
-    app.all(
-      /^\/case\/.+\/response\/(?![\d]+\/receiver|confirmation|full-admission|partial-admission|counter-claim|receipt).*$/,
-      CountyCourtJudgmentRequestedGuard.requestHandler
-    )
+    app.all(/^\/case\/.+\/response\/.*$/, CountyCourtJudgmentRequestedGuard.requestHandler)
     app.all(
       /^\/case\/.+\/response\/(?![\d]+\/receiver|confirmation|receipt).*$/,
       DraftMiddleware.requestHandler('response', (value: any): ResponseDraft => {
--- a/src/main/features/response/index.ts
+++ b/src/main/features/response/index.ts
@@ -10,6 +10,7 @@ import { AlreadyRespondedGuard } from 'response/guards/alreadyRespondedGuard'
 import { ClaimMiddleware } from 'app/claims/claimMiddleware'
 import { DraftMiddleware } from 'common/draft/draftMiddleware'
 import { ResponseDraft } from 'response/draft/responseDraft'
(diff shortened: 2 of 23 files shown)


src/main/features/response/guards/countyCourtJudgmentRequestedGuard.ts
```diff
@@ -0,0 +1,17 @@
+import * as express from 'express'
```
timja-kainos: this is unexpected...
This functionality is scheduled for next sprint in a story which needs some design work:
https://tools.hmcts.net/jira/browse/ROC-2485

src/test/features/response/routes/defence-options.ts
```diff
@@ -6,6 +6,7 @@ import { attachDefaultHooks } from '../../../routes/hooks'
 import '../../../routes/expectations'
 import { checkAuthorizationGuards } from './checks/authorization-check'
 import { checkAlreadySubmittedGuard } from './checks/already-submitted-check'
+import { checkCountyCourtJudgmentRequestedGuardGuard } from './checks/ccj-requested-check'
```
geek-soft: I cannot find `ccj-requested-check` anywhere in the committed files :(

src/test/features/response/routes/defence-options.ts
```diff
@@ -35,6 +36,7 @@ describe('Defendant response: defence options page', () => {
       })
 
       checkAlreadySubmittedGuard(app, 'get', defenceOptionsPage)
+      checkCountyCourtJudgmentRequestedGuardGuard(app, 'get', defenceOptionsPage)
```
geek-soft: We should also be testing that query is applied to POST requests.

src/main/features/response/index.ts
```diff
@@ -34,6 +35,10 @@ export class Feature {
       AlreadyRespondedGuard.requestHandler
     )
     app.all(
+      /^\/case\/.+\/response\/(?![\d]+\/receiver|confirmation|full-admission|partial-admission|counter-claim|receipt).*$/,
```
geek-soft: For me it looks like this regex matches all paths starting with `/case/<something>/response` except `/case/<something>/response/<some-digits>/receiver`. IF only linking receiver is supported to be excluded then there is no point in having `confirmation|full-admission|partial-admission|counter-claim|receipt` as a part of regex.

src/main/features/response/guards/countyCourtJudgmentRequestedGuard.ts
geek-soft: Worth checking with business how not having it impacts planned release. I do not mind this temporary solution in place if that would let us make feature available to the users sooner.

timja-kainos: Clement Neveu [4:06 PM] 
@Tim I think we should be able to release without it: The judgment will be recorded on Caseman and the support team will therefore not process the Defence and send it out to the claimant (edited)

src/main/features/response/index.ts
timja-kainos: the receiver isn't even under case...

src/main/features/response/guards/countyCourtJudgmentRequestedGuard.ts
```diff
@@ -0,0 +1,17 @@
...
+import User from 'idam/user'
+const logger = require('@hmcts/nodejs-logging').getLogger('response/guards/countyCourtJudgmentRequestedGuard')
+
+export default class CountyCourtJudgmentRequestedGuard {
+
+  static requestHandler (req: express.Request, res: express.Response, next: express.NextFunction) {
+    const user: User = res.locals.user
+    if (user.claim.countyCourtJudgmentRequestedAt) {
+      logger.warn('State guard: CCJ already requested - redirecting to dashboard')
+      return res.redirect(DashboardPaths.dashboardPage.uri)
```
timja-kainos: Need to check whether the dashboard is enough messaging, this isn't someone doing anything wrong its something that could easily happen.
Linking of a defendant to a claim is where I am most worried about this...

I would also suggest waiting till the next sprint so that we can finish everything in the current sprint...

src/main/features/response/guards/countyCourtJudgmentRequestedGuard.ts
geek-soft: However defence is then recorded on the case which puts it in inconsistent state. I would pretty much prefer to avoid it, despite the fact risk might not be huge.

src/main/features/response/guards/countyCourtJudgmentRequestedGuard.ts
geek-soft: Agree, that is not perfect but protects us from inconsistency in case data. I would personally consider holding on with the release without this in place.

src/test/features/response/routes/defence-options.ts
yarpo: In that case I'd also update tests for `checkAlreadySubmittedGuard`. Do you agree?

src/test/features/response/routes/defence-options.ts
yarpo: You're right. Added.

src/main/features/response/guards/countyCourtJudgmentRequestedGuard.ts
```diff
@@ -0,0 +1,17 @@
+import * as express from 'express'
+import { Paths as DashboardPaths } from 'dashboard/paths'
+import User from 'idam/user'
+const logger = require('@hmcts/nodejs-logging').getLogger('response/guards/countyCourtJudgmentRequestedGuard')
+
+export default class CountyCourtJudgmentRequestedGuard {
```
geek-soft: Please change it to not be a default export.

src/test/features/response/routes/check-and-send.ts
```diff
@@ -6,6 +6,7 @@ import { attachDefaultHooks } from '../../../routes/hooks'
 import '../../../routes/expectations'
 import { checkAuthorizationGuards } from './checks/authorization-check'
 import { checkAlreadySubmittedGuard } from './checks/already-submitted-check'
+import { checkCountyCourtJudgmentRequestedGuardGuard } from './checks/ccj-requested-check'
```
geek-soft: It should be `checkCountyCourtJudgmentRequestedGuard` not `checkCountyCourtJudgmentRequestedGuardGuard`. Remove duplicate `Guard` from the name.

src/main/features/response/guards/countyCourtJudgmentRequestedGuard.ts
yarpo: ok

src/test/features/response/routes/checks/ccj-requested-check.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import '../../../../routes/expectations'
+
+import * as claimStoreServiceMock from '../../../../http-mocks/claim-store'
+import { Paths } from 'dashboard/paths'
+
+const cookieName: string = config.get<string>('session.cookieName')
+
+export function checkCountyCourtJudgmentRequestedGuardGuard (app: any, method: string, pagePath: string) {
+  it(`for ${method} should redirect to your dashboard page when defendant has already responded`, async () => {
```
geek-soft: That is misleading. The simulated scenario is when CCJ was requested not that defendant already responded. It should be `for ${method} should redirect to your dashboard page when claimant has already requested CCJ`

src/test/features/response/routes/counter-claim.ts
```diff
@@ -28,11 +30,14 @@ describe('Defendant response: counter claim page', () => {
         idamServiceMock.resolveRetrieveUserFor(1, 'cmc-private-beta', 'defendant')
       })
 
+      checkCountyCourtJudgmentRequestedGuardGuard(app, 'get', pagePath)
+      checkCountyCourtJudgmentRequestedGuardGuard(app, 'post', pagePath)
```
geek-soft: That is not right place for it. I do not think testing POST requests in `'on GET'` block is what we tend to do.

yarpo: My mistake. This route has no POST handler at all. Removed testing POST then

src/test/features/response/routes/check-and-send.ts
yarpo: well spotted

geek-soft:
Guard test is missing for `features/response/routes/response-receipt.ts`.

geek-soft:
LGTM now 👍

PR #98, opened 2017-10-06
ROC-2480 Fix calculating 10,000 amount for handoff

timja-kainos:


timja-kainos:
There was no need to check this on this page, the user doesn't enter any value, it was the only place checking it wrong too

src/main/features/claim/routes/total.ts
```diff
@@ -19,6 +18,5 @@ export default express.Router()
       .catch(next)
   })
   .post(Paths.totalPage.uri, (req: express.Request, res: express.Response) => {
-    ClaimValidator.claimAmount(req.body.totalAmount)
```
geek-soft: There is the following on the `total.njk`. We should remove it since no longer in use.

```
<input type="hidden" name="totalAmount" value="{{ interestTotal.totalAmount }}">
```

geek-soft: Deleted it.

PR #100, opened 2017-10-06
Deserialise optional defendant email address to udnefined if not provided

geek-soft:
### JIRA link ###

None

### Change description ###

This is to deserialise optional defendant email address to udnefined if value is not provided e.g. is blank. Followup PR will be raised for the backend to migrate existing cases.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/test/app/utils/stringUtils.ts
+++ b/src/test/app/utils/stringUtils.ts
@@ -1,0 +1,28 @@
+/* tslint:disable:no-unused-expression */
+import { expect } from 'chai'
+
+import StringUtils from 'utils/stringUtils'
+
+describe('StringUtils', () => {
+  describe('trimToUndefined', () => {
+    it('should return undefined if value is undefined', () => {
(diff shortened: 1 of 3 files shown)


src/test/app/utils/stringUtils.ts
```diff
@@ -0,0 +1,28 @@
+/* tslint:disable:no-unused-expression */
+import { expect } from 'chai'
+
+import StringUtils from 'utils/stringUtils'
+
+describe('StringUtils', () => {
+  describe('trimToUndefined', () => {
+    it('should return undefined if value is undefined', () => {
```
yarpo: what about `null`?

geek-soft: Decided to code against undefined only because we tend to avoid nulls. Can add this on assumption that is candidate library for sharing with others.

timja-kainos: no nulls

PR #101, opened 2017-10-06
ROC-2469 Disable unknown external refernce usage'

dharmendrak:


src/main/features/claim/routes/confirmation.ts
```diff
@@ -3,12 +3,14 @@ import { Paths } from 'claim/paths'
 
 import ClaimStoreClient from 'claims/claimStoreClient'
 import Claim from 'claims/models/claim'
+import OwnershipChecks from 'app/auth/ownershipChecks'
```
yarpo: @geek-soft didn't we decide we don't use `default exports` in TypeScript?

yarpo: Why is this not a guard?

src/main/features/claim/routes/confirmation.ts
```diff
@@ -3,12 +3,14 @@ import { Paths } from 'claim/paths'
...
 
 export default express.Router()
   .get(Paths.confirmationPage.uri, async (req: express.Request, res: express.Response, next: express.NextFunction) => {
     const { externalId } = req.params
     try {
       const claim: Claim = await ClaimStoreClient.retrieveByExternalId(externalId)
+      OwnershipChecks.checkClaimOwner(res.locals.user, claim)
```
timja-kainos: This is not the right approach.
`ClaimStoreClient.retrieveByExternalId` should have the check and there need's to be a forbidden page explaining what happened

src/main/features/claim/routes/confirmation.ts
geek-soft: You are right @yarpo , we tend to not use defaults. Please change it @dharmendrak

src/main/features/claim/routes/confirmation.ts
geek-soft: I would just slightly change the approach to not add the ownership check logic to `ClaimStoreClient.retrieveByExternalId` and creating `ClaimStoreService` instead which delegates data retrieval to generic client and throws an error when ownership check failed.

That would keep HTTP client simple with one responsibility only and in sync with the API which does not have any ownership logic.

src/main/features/claim/routes/confirmation.ts
```diff
@@ -8,7 +8,7 @@ export default express.Router()
   .get(Paths.confirmationPage.uri, async (req: express.Request, res: express.Response, next: express.NextFunction) => {
     const { externalId } = req.params
     try {
-      const claim: Claim = await ClaimStoreClient.retrieveByExternalId(externalId)
+      const claim: Claim = await ClaimStoreClient.retrieveByExternalId(externalId, res.locals.user.id)
```
timja-kainos: might be worth some tests for all this

src/main/app/claims/claimStoreClient.ts
```diff
@@ -69,7 +69,11 @@ export default class ClaimStoreClient {
       .get(`${claimStoreApiUrl}/${externalId}`)
       .then(claim => {
         if (claim) {
-          return new Claim().deserialize(claim)
+          if (userId !== claim.submitterId) {
+            throw new Error('You are not allowed to access this resource')
```
rlewan: Would be nice if this was e.g. a `ForbiddenError` which would map to a forbidden page.

src/main/app/claims/claimStoreClient.ts
```diff
@@ -69,7 +70,11 @@ export default class ClaimStoreClient {
       .get(`${claimStoreApiUrl}/${externalId}`)
       .then(claim => {
         if (claim) {
-          return new Claim().deserialize(claim)
+          if (userId !== claim.submitterId) {
```
geek-soft: That breaks response journey! Defendant is unable to retrieve claim by external ID

timja-kainos: oops, :(

timja-kainos: how did the checks pass if it breaks them?

pragnesh-patel: I was thinking how the checks passed. How???

timja-kainos: @rlewan is investigating

PR #105, opened 2017-10-10
ROC-2062: Fail view rendering when data is undefined

geek-soft:
### JIRA link ###

https://tools.hmcts.net/jira/browse/ROC-2062

### Change description ###

This PR is to make Nunjucks fail on missing data. That should help us avoid problems where somebody refactored template or data model where as a result view is rendered without all necessary elements. After introducing this change view rendering will fail fast with an exception which will give us chance to spot the problem straight away.

TODO:
- find all the places where template accepts undefined data and make optional data explicit
- fix route tests & a11y tests

```
[ ] Yes
[x] No
```


src/test/routes/hooks.ts
```diff
@@ -1,11 +1,18 @@
...
 
 import * as idamServiceMock from '../http-mocks/idam'
 
-export function attachDefaultHooks () {
+export function attachDefaultHooks (app?: express.Express) {
   let retrieveServiceTokenMock: any
 
+  before(() => {
+    if (app) {
+      app.locals.csrf = 'dummy-token'
```
geek-soft: Significant amount of route tests started failing after making rendering more strict due to the fact that CSRF token wasn't there. Enabling CSRF module for tests is still not convenient for tests so this sets dummy token instead just to make views render.

src/test/a11y/mocks.ts
```diff
@@ -1,197 +1,33 @@
...
-  draft.defendant.email = new Email('defendant@hmcts.com')
-  individualDefendant.address.postcode = 'postcode'
-  individualDefendant.address.line1 = 'line1'
-  individualDefendant.dateOfBirth = new DateOfBirth(true, new LocalDate(1980, 1, 1))
-  draft.defendant.mobilePhone = new MobilePhone('07873738765')
-  draft.defendant.partyDetails = new PartyDetails()
-  draft.defendant.partyDetails.address.postcode = 'postcode'
-  draft.defendant.partyDetails.address.line1 = 'line1'
-  draft.interestDate.date = new LocalDate(2017, 7, 21)
+idamServiceMock.resolveRetrieveUserFor(1, 'cmc-private-beta', 'letter-holder', 'claimant', 'defendant').persist()
```
geek-soft: A `persist()` method keeps mock in place forever which is different to the ordinary behaviour where mock is being removed after being used for the first time.

geek-soft:
Will need to update version of `@hmcts/cmc-common-frontend` when new version is released.

src/test/a11y/a11y.ts
```diff
@@ -20,9 +22,10 @@ function check (url: string): void {
   describe(`Page ${url}`, () => {
 
     it('should have no accessibility errors', (done) => {
-      ensurePageCallWillSucceed(url)
+      const urlWithParams = `${url}?jwt=ABC&ref=000MC000`
```
timja-kainos: this looks wrong, and this hack with jwt= won't work when the oauth stuff is turned on

PR #118, opened 2017-10-17
ROC-2069: Update check-and-send page after changing response options

rattur:
### JIRA link ###

https://tools.hmcts.net/jira/browse/ROC-2069

### Change description ###

Updated check and send page to show do you owe money claimed section

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/main/features/response/views/check-and-send.njk
+++ b/src/main/features/response/views/check-and-send.njk
@@ -27,6 +27,38 @@
 
     {{ tableEnd() }}
 
+    {{ tableStart("moneyOwed", "Do you owe the money claimed?") }}
+
+    {{ row('Do you owe the money claimed?', t(draft.response.type.displayValue) | capitalize, paths.responseTypePage.evaluateUri({ externalId: claim.externalId })) }}
+
+    {% if draft.response.type.value === 'OWE_SOME_PAID_NONE' %}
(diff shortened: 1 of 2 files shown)


src/main/features/response/views/check-and-send.njk
```diff
@@ -27,6 +27,38 @@
...
+
+    {{ row('Do you owe the money claimed?', t(draft.response.type.displayValue) | capitalize, paths.responseTypePage.evaluateUri({ externalId: claim.externalId })) }}
+
+    {% if draft.response.type.value === 'OWE_SOME_PAID_NONE' %}
+      {% if draft.rejectPartOfClaim.option === 'paidWhatBelievedWasOwed' %}
+        {{ row('Why are you rejecting part of this claim?', t('I’ve paid what I believe I owe') | capitalize, paths.defenceRejectPartOfClaimPage.evaluateUri({ externalId: claim.externalId })) }}
+      {% endif %}
+
+      {% if draft.rejectPartOfClaim.option === 'amountTooHigh' %}
+        {{ row('Why are you rejecting part of this claim?', t('The claim amount is too much') | capitalize, paths.defenceRejectPartOfClaimPage.evaluateUri({ externalId: claim.externalId })) }}
```
geek-soft: A `Why are you rejecting part of this claim?` is duplicated. Better option would be having one row macro call and pass variable with the value which is evaluated depending on the option selected.

PR #122, opened 2017-10-17
Fix to change to task list builder flow and update check and send

rattur:
### JIRA link ###

https://tools.hmcts.net/jira/browse/ROC-1023

### Change description ###

Change task list builder and check and send template

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/main/features/response/helpers/taskListBuilder.ts
+++ b/src/main/features/response/helpers/taskListBuilder.ts
@@ -34,15 +34,14 @@ export class TaskListBuilder {
       OweMoneyTask.isCompleted(draft)))
 
     if (draft.requireHowMuchPaid()) {
-      tasks.push(new TaskListItem('How much have you paid the claimant?', Paths.defendantHowMuchPaid.
-        evaluateUri({ externalId: externalId }),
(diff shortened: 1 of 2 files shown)


geek-soft:
@rattur please add story ID when you merge so we can track it in the commit history.

src/main/features/response/helpers/taskListBuilder.ts
```diff
@@ -34,15 +34,14 @@ export class TaskListBuilder {
       OweMoneyTask.isCompleted(draft)))
 
     if (draft.requireHowMuchPaid()) {
```
geek-soft: This reads wired. When `requireHowMuchPaid()` condition evaluates to true we add `'How much money do you believe you owe?'` to the task list. In turn when `requireHowMuchOwed()` evaluates to true we add `'How much have you paid the claimant?'` to the task list. My first impression is that `much paid` and `much owed` have swapped place.

rattur: Yes thats right, it is a bit confusing, which is why originally they were in that order, then having had a look at the updated prototype they needed to be swapped

rlewan: Yeah, this will keep confusing people each time anybody opens that bit of code. We should come up with better naming.

PR #124, opened 2017-10-18
Remove number assumptions on defendant type

timja-kainos:
The type for the IDAM identifier was an implementation detail and should
not have been assumed.

The type is changing soon, to enable testing of the new solution we need
to remove our assumption on the type

### DIFF
--- a/src/main/app/claims/claimStoreClient.ts
+++ b/src/main/app/claims/claimStoreClient.ts
@@ -32,7 +32,7 @@ export default class ClaimStoreClient {
     })
   }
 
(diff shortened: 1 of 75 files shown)


src/main/app/claims/claimStoreClient.ts
```diff
@@ -32,7 +32,7 @@ export default class ClaimStoreClient {
     })
   }
 
-  static retrieveByClaimantId (claimantId: number): Promise<Claim[]> {
+  static retrieveByClaimantId (claimantId: string): Promise<Claim[]> {
```
yarpo: why string?

PR #129, opened 2017-10-20
ROC-2446 Fix dashboard on mobile

timja-kainos:


src/main/features/dashboard/views/index.njk
```diff
@@ -4,82 +4,67 @@
...
-                    t('You requested a County Court Judgment on %s. We will contact you within 5 days.',
-                      { postProcess: 'sprintf', sprintf: [ claim.countyCourtJudgmentRequestedAt | date ]})
-                  }}
-                {% else %}
-                  {{ t('Your claim is being processed. We will email you to explain what to do next.') }}
-                {% endif %}
-              </td>
+              <td class="mobile-hide">{{ status }}</td>
             </tr>
+            <tr><td colspan="3" class="mobile-table-status">{{ t('Status: ') + status }}</td></tr>
```
rlewan: Shouldn't `status` be translated too?

rlewan: Oh it is translated earlier, ignore.

rlewan: Or -- what do you think about translating the status here instead of in the conditional block above?

timja-kainos: can't really because some stuff shouldn't get translated and each status content is quite different in regards to placeholders

rlewan: 👍

PR #138, opened 2017-10-23
Workaround for unaligned response type models

rlewan:
### Change description ###

`OWE_ALL_PAID_ALL` response type has been split into response subtypes. But Claim Store still relies on it to compose a staff notification. This change is to provide a workaround that will fix the broken functionality. Frontend and backend models will be aligned properly as a part of finishing partial admission stories.

```
[ ] Yes
[X] No
```


src/main/app/claims/responseModelConverter.ts
```diff
@@ -25,7 +27,7 @@ export class ResponseModelConverter {
       )
     }
     return new ResponseData(
-      responseDraft.response.type.value,
+      this.inferResponseType(responseDraft),
```
yarpo: can we please add in the next PR unit tests for `inferResponseType` logic?

PR #142, opened 2017-10-25
ROC-2469 Prevent wrong user accessing page

timja-kainos:
### JIRA link ###
https://tools.hmcts.net/jira/browse/ROC-2469


src/test/features/offer/routes/your-offer.ts
```diff
@@ -76,7 +76,7 @@ describe('Offer page', () => {
           it('should redirect to offer confirmation page', async () => {
             claimStoreServiceMock.resolveRetrieveClaimByExternalId()
             claimStoreServiceMock.resolveSaveOffer()
-            const futureDate = moment().add(4, 'month')
+            const futureDate = moment().day(1).add(4, 'month')
```
geek-soft: That looks suspicious. Previous method should take care of initiating date in the future. I think the problem was that correct second argument is `months` not `month` so we should have `.add(4, 'months')`.

I also dislike that `day(int)` is quite enigmatic, `add(int, string)` is a bit more readable IMO.

PR #154, opened 2017-10-27
ROC-2475: Add UI to view of offer to settle the claim

comanirudha:
### JIRA link ###
https://tools.hmcts.net/jira/browse/ROC-2475

### Change description ###

Claimant view of offer to settle the claim

```
[ ] Yes
[ x] No
```


src/main/features/offer/routes/defendant-response.ts
```diff
@@ -0,0 +1,44 @@
...
+    form: form,
+    claim : res.locals.user.claim,
+    offer: getOffer(res.locals.user.claim)
+  })
+}
+
+export default express.Router()
+  .get(
+    Paths.defendantResponsePage.uri,
+    OfferGuard.requestHandler,
```
geek-soft: It does not make sense to have it here since guard redirects to conformation page. See below logic extracted from the guard.

```
logger.warn(`Claim ${claim.claimNumber} not eligible for an offer - redirecting to confirmation page`)
res.redirect(Paths.confirmationPage.evaluateUri({ externalId: user.claim.externalId }))
```

Please also use `GuardFactory.create` in the guard so that we have single class to mock and common mechanism of writing guards.

src/main/features/offer/routes/defendant-response.ts
```diff
@@ -0,0 +1,44 @@
...
+import { DefendantResponse } from 'offer/form/models/defendantResponse'
+import { ErrorHandling } from 'common/errorHandling'
+import User from 'idam/user'
+import { OfferGuard } from 'offer/guards/offerGuard'
+
+function getOffer (claim: any): any {
+  if (!claim.settlement || !claim.settlement.partyStatements) {
+    return undefined
+  }
+  return claim.settlement.partyStatements[0].offer
```
geek-soft: This should return last offer not the first one since the page will be reused for responding to more then one defendant offer. Implement something similar to `uk.gov.hmcts.cmc.claimstore.models.offers.Settlement#getLastStatement`

src/main/features/offer/views/defendant-response.njk
```diff
@@ -0,0 +1,69 @@
+{% extends "layout.njk" %}
+{% from "form.njk" import csrfProtection, submitButton, errorSummary, radioGroup %}
+{% from "date.njk" import dateInput %}
+
+{% set title = t('Money claim') %}
+
+{% block content %}
+  <div class="grid-row">
+    <div class="column-two-thirds">
+      <h1 class="heading-large form-hint">{{ t("Defendant's response") }} </h1>
```
geek-soft: Please use single quotes and correct apostrophe char.

src/main/features/offer/views/defendant-response.njk
```diff
@@ -0,0 +1,69 @@
...
+      <form novalidate method="post">
+        {{ csrfProtection(csrf) }}
+        <div class="description-block">
+          <h2 class="heading-large"> 
+          {{ t('%s’s offer',{
+              postProcess: 'sprintf',
+              sprintf: [claim.claimData.defendant.name]
+            }) }}</h2>
+        </div>
+        <table class="claim-dashboard">
```
geek-soft: That does not make sense to have it here. Please remove or reimplement for an offer table.

src/main/features/offer/views/defendant-response.njk
```diff
@@ -0,0 +1,69 @@
...
+          <tr class="claims default mobile-table">
```
geek-soft: The same here, why `claims` class is used here?

geek-soft: What `default` class is doing here for us? Is it declared anywhere?

src/main/features/offer/views/defendant-response.njk
```diff
@@ -0,0 +1,69 @@
...
+              <td>{{ offer.completionDate | date }}</td>
+            </tr>
+        </table>
+
+        <h2 class="heading-medium"> {{ t('Do you accept the offer?') }} </h2>
+
+        {{
+        radioGroup(
+          form = form,
+          label = 'Yes',
```
geek-soft: That does not look right. That is the label for fieldset

src/test/features/offer/routes/defendant-response.ts
```diff
@@ -0,0 +1,97 @@
...
+            await request(app)
+              .post(defendantResponsePage)
+              .set('Cookie', `${cookieName}=ABC`)
+              .send({})
+              .expect(res => expect(res).to.be.serverError.withText('Error'))
+          })
+        })
+
+        context('when form is valid', async () => {
+          it('should display defendantResponsePage', async () => {
```
geek-soft: We tend to say `should render page`. No need to be specific since that make maintenance harder in case page is renamed.

src/main/features/offer/paths.ts
```diff
@@ -6,4 +6,5 @@ export class Paths {
   static readonly offerPage = new RoutablePath(`${offerPath}/your-offer`)
   static readonly offerConfirmationPage = new RoutablePath(`${offerPath}/offer-confirmation`)
   static readonly settleOutOfCourtPage = new RoutablePath(`${offerPath}/settle-out-of-court`)
+  static readonly defendantResponsePage = new RoutablePath(`${offerPath}/defendant-response`)
```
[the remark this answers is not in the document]
geek-soft: What about `/offer/summary` or `/offer/response`?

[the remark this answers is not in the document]
comanirudha: /offer/response sounds right, I will change it.

src/main/app/claims/models/partyStatement.ts
```diff
@@ -0,0 +1,24 @@
+
+import { Serializable } from 'models/serializable'
+import { Offer } from 'claims/models/offer'
+
+export class PartyStatement implements Serializable<PartyStatement> {
+  type: string
+  madeBy: string
+  offer: Offer
```
rlewan: This field is optional on the backend side, should be optional here as well.

src/main/features/offer/form/models/defendantResponse.ts
```diff
@@ -0,0 +1,30 @@
+import { IsDefined, IsIn } from 'class-validator'
+import { Serializable } from 'models/serializable'
+
+export class ValidationErrors {
+  static readonly OPTION_REQUIRED: string = 'Choose option: yes or no or make an offer'
+}
+
+export class DefendantResponse implements Serializable<DefendantResponse> {
+  @IsDefined({ message: ValidationErrors.OPTION_REQUIRED })
+  @IsIn(['yes', 'no', 'makeAnCounterOffer'], { message: ValidationErrors.OPTION_REQUIRED })
```
rlewan: These values should be defined in a constants class , e.g. like in [`EvidenceType`](https://github.com/hmcts/cmc-citizen-frontend/blob/master/src/main/features/response/form/models/evidenceType.ts). This constants class should be also exported as a nunjucks global, see [here](https://github.com/hmcts/cmc-citizen-frontend/blob/master/src/main/modules/nunjucks/index.ts#L78) for details.

src/main/features/offer/routes/defendant-response.ts
```diff
@@ -0,0 +1,44 @@
...
+import { Paths } from 'offer/paths'
+import { Form } from 'forms/form'
+import { DefendantResponse } from 'offer/form/models/defendantResponse'
+import { ErrorHandling } from 'common/errorHandling'
+import User from 'idam/user'
+import { OfferGuard } from 'offer/guards/offerGuard'
+
+function getOffer (claim: any): any {
+  if (!claim.settlement || !claim.settlement.partyStatements) {
+    return undefined
```
rlewan: You are returning `undefined` here, but assuming that the offer will be present on the page. Please either make an assertion here that an offer is present when this route is accessed or add logic to the page which correctly handles missing offer. I would lean towards the former.

src/main/app/claims/models/offer.ts
```diff
@@ -7,7 +7,7 @@ export class Offer implements Serializable<Offer> {
   content: string
   completionDate: Moment
 
-  constructor (content: string, completionDate: Moment) {
+  constructor (content?: string, completionDate?: Moment) {
```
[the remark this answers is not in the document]
comanirudha: The current pattern is to use X(). deserialise, may be we need a overall code refactoring which can be done as a separate story.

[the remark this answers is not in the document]
geek-soft: Definitely not for now, I just pointed out the side effect of `new X().deserialise` pattern which I do not like.

src/main/features/offer/form/models/defendantResponse.ts
```diff
@@ -0,0 +1,30 @@
...
+import { Serializable } from 'models/serializable'
+import { StatementType } from './statementType'
+export class ValidationErrors {
+  static readonly OPTION_REQUIRED: string = 'Choose option: yes or no or make an offer'
+}
+
+export class DefendantResponse implements Serializable<DefendantResponse> {
+  @IsDefined({ message: ValidationErrors.OPTION_REQUIRED })
+  @IsIn([StatementType.ACCEPTATION.value, StatementType.REJECTION.value, StatementType.OFFER.value], { message: ValidationErrors.OPTION_REQUIRED })
+  option?: string
```
geek-soft: Please add validation unit test similar to `features/response/form/models/rejectPartOfClaim.ts` for example.

src/main/features/offer/routes/defendant-response.ts
geek-soft: This has not been addressed yet!!!

src/main/features/offer/routes/response.ts
```diff
@@ -0,0 +1,56 @@
...
+import { Paths } from 'offer/paths'
+import { Form } from 'forms/form'
+import { DefendantResponse } from 'offer/form/models/defendantResponse'
+import { ErrorHandling } from 'common/errorHandling'
+import User from 'idam/user'
+import { OfferGuard } from 'offer/guards/offerGuard'
+
+function getPartyStatement (claim: any): any[] {
+  if (!claim.settlement || !claim.settlement.partyStatements) {
+    throw Error('Something went wrong, No offer present')
```
geek-soft: This is not desired solution. Having this would mean that when citizen hits the page URI at a time there is no offer they will get 500 error. We wanna handle this scenario more gracefully so the possible solutions are:

- having business error page saying offer has not been made by defendant yet or more generic one saying an action is not available at this stage (preferred)
- redirect silently to dashboard without error message

src/main/features/offer/routes/response.ts
```diff
@@ -0,0 +1,56 @@
+import * as express from 'express'
+import { FormValidator } from 'forms/validation/formValidator'
+import { Paths } from 'offer/paths'
+import { Form } from 'forms/form'
+import { DefendantResponse } from 'offer/form/models/defendantResponse'
+import { ErrorHandling } from 'common/errorHandling'
+import User from 'idam/user'
+import { OfferGuard } from 'offer/guards/offerGuard'
+
+function getPartyStatement (claim: any): any[] {
```
geek-soft: Please declare return type correctly. We aim to avoid any. In your case it is `PartyStatement[]`

src/main/app/claims/models/partyStatement.ts
```diff
@@ -0,0 +1,24 @@
+
+import { Serializable } from 'models/serializable'
+import { Offer } from 'claims/models/offer'
+
+export class PartyStatement implements Serializable<PartyStatement> {
+  type?: string
+  madeBy?: string
+  offer?: Offer
```
rlewan: I my previous comment I meant that only `offer` is an optional field. Other two should be mandatory.

src/main/features/offer/views/response.njk
```diff
@@ -0,0 +1,68 @@
...
+          name = 'option',
+          options = [
+            {
+              bold: false,
+              label: 'Yes',
+              hint: t('You can make a legal agreement between you and %s.',
+                { postProcess: 'sprintf',
+                sprintf:[claim.claimData.defendant.name]
+                }),
+              value: 'StatementType.ACCEPTATION.value',
```
rlewan: You are passing the `'StatementType.ACCEPTATION.value'` string here, not accessing `value` property of `StatementType.ACCEPTATION`. Same in other places below.

src/test/http-mocks/claim-store.ts
```diff
@@ -66,6 +66,10 @@ export const sampleClaimObj = {
     defendantDateOfBirth: '1990-11-01',
     paidAmount: 2,
     paymentOption: 'IMMEDIATELY'
+  },
+  settlement: {
+    partyStatements:
+        [ { type: 'OFFER', madeBy: 'DEFENDANT' , offer: { content: 'offer text', completionDate: '2017-08-08'}}]
```
rlewan: Can we use the new constants here as well?

src/main/features/offer/form/models/defendantResponse.ts
comanirudha: we already have some router tests which test validation.

src/main/features/offer/routes/response.ts
comanirudha: the use can only see this page if they have offer, if they don't the link is usually disabled. This if only safeguards against directly entering url in the browser.

src/main/features/offer/form/models/defendantResponse.ts
comanirudha: done

timja-kainos:
conflicts

src/main/app/claims/models/partyStatement.ts
```diff
@@ -0,0 +1,24 @@
...
+    this.type = type
+    this.madeBy = madeBy
+    this.offer = offer
+  }
+
+  deserialize (input: any): PartyStatement {
+    if (input) {
+      this.type = input.type
+      this.madeBy = input.madeBy
+      this.offer = new Offer().deserialize(input.offer)
```
geek-soft: If offer is optional I think we should have a condition here so that when undefined was returned from backend we have `PartyStatement.undefined `. I think about something like:

```
if (input.offer) {
  this.offer = new Offer().deserialize(input.offer)
}
```

src/main/features/offer/form/models/defendantResponse.ts
geek-soft: Just to clarify route test is not a place to test validation. I roughly test that validation is enabled but not that validation works fine. In route tests we test one scenario but here we have 3 possible options and we need to test all are handled properly.

src/main/features/offer/form/models/offer.ts
```diff
@@ -17,24 +17,27 @@ export class Offer implements Serializable<Offer> {
...
   @IsDefined({ message: ValidationErrors.DATE_REQUIRED })
   @IsValidLocalDate({ message: ValidationErrors.DATE_NOT_VALID })
   @IsFutureDate({ message: ValidationErrors.FUTURE_DATE })
-  completionDate?: LocalDate
+  completionDate: LocalDate
 
-  constructor (offerText?: string, completionDate?: LocalDate) {
-    this.offerText = offerText
-    this.completionDate = completionDate
+  static create (offerText: string, completionDate: LocalDate) {
```
geek-soft: What is the purpose of this? Why did you introduce it over regular constructor? If there is no significant advantage of this I would suggest reverting back to constructor as this is the pattern / expected way of instantiating classes.

src/main/features/offer/form/models/offer.ts
```diff
@@ -17,24 +17,27 @@ export class Offer implements Serializable<Offer> {
...
+    const offer = new Offer()
+    if (offerText || completionDate) {
```
geek-soft: What is the purpose of this? Populating fields if any of two arguments is provided looks wired at least. This also in conflict with the facts that both fields are marked as mandatory but here it is possible to instantiate only one field.

src/main/features/offer/routes/response.ts
```diff
@@ -0,0 +1,65 @@
...
+import User from 'idam/user'
+import { OfferGuard } from 'offer/guards/offerGuard'
+import { StatementType } from 'offer/form/models/statementType'
+import { MadeBy } from 'offer/form/models/madeBy'
+
+function redirectToDashBoardPage (res: express.Response) {
+  const user: User = res.locals.user
+  res.redirect(DashboardPaths.claimantPage.evaluateUri({ externalId: user.claim.externalId }))
+}
+function getPartyStatement (claim: any, res: express.Response): PartyStatement[] {
```
geek-soft: Correct type, claim is of type of Claim, not any. Lack of types is error prone.

src/main/features/offer/routes/response.ts
```diff
@@ -0,0 +1,65 @@
...
+    redirectToDashBoardPage(res)
+  }
+
+  return claim.settlement.partyStatements.map(partyStatement => {
+    if (partyStatement.type === StatementType.OFFER.value && partyStatement.madeBy === MadeBy.DEFENDANT.value) {
+      return partyStatement
+    }
+  })
+}
+function getOffer (claim: any, res: express.Response): Offer {
```
geek-soft: Correct type, remove `res` argument.

src/main/features/offer/routes/response.ts
```diff
@@ -0,0 +1,65 @@
...
+import { StatementType } from 'offer/form/models/statementType'
+import { MadeBy } from 'offer/form/models/madeBy'
+
+function redirectToDashBoardPage (res: express.Response) {
+  const user: User = res.locals.user
+  res.redirect(DashboardPaths.claimantPage.evaluateUri({ externalId: user.claim.externalId }))
+}
+function getPartyStatement (claim: any, res: express.Response): PartyStatement[] {
+  if (!claim.settlement || !claim.settlement.partyStatements) {
+    redirectToDashBoardPage(res)
```
geek-soft: This is side effect. A `getPartyStatement` should only return value, not do redirects. It violates CQRS principle.

Do that logic in the place which retrieved the value from `getPartyStatement`. If undefined you can make redirect. Also guard could be possible solution as probably no point even starting processing in that router when settlement is not available.

src/main/features/offer/routes/response.ts
geek-soft: Not a valid place to make redirect. See my comment above.

src/test/features/offer/form/models/defendantResponse.ts
```diff
@@ -0,0 +1,57 @@
...
+      expect(new DefendantResponse().deserialize({ option: StatementType.ACCEPTATION.value})).to.deep.equal(new DefendantResponse(StatementType.ACCEPTATION.value))
+    })
+  })
+
+  describe('validation', () => {
+    const validator: Validator = new Validator()
+
+    describe('should reject when', () => {
+      it('undefined option', () => {
+        const errors = validator.validateSync(new DefendantResponse())
```
geek-soft: Please pass undefined explicitly, in case default behaviour of the constructor is changed.

Plase also test that it is rejected when non valid option is passed, e.g. `new DefendantResponse('unknown-value')`

src/test/features/offer/form/models/offer.ts
```diff
@@ -69,10 +69,10 @@ describe('Offer', () => {
 
     describe('should accept when', () => {
       it('offer text and future date', () => {
-        const futureDate = moment().add(10, 'days')
-        const date = new LocalDate(futureDate.year(), futureDate.month(), futureDate.day())
-        const errors = validator.validateSync(new Offer('offer text', date))
-        expect(errors.length).to.equal(1)
+        const futureDate = moment().add(10, 'month')
```
geek-soft: Why did you have to make that change? It looks not related, smells like something bad is happening here.

src/main/features/offer/form/models/defendantResponse.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  constructor (option?: StatementType) {
+    this.option = option
+  }
+
+  static fromObject (value?: any): DefendantResponse {
+    if (!value) {
+      return value
+    }
+    // tslint:disable-next-line:no-console
```
timja-kainos: umm

PR #156, opened 2017-10-29
Node process per CPU core

rlewan:
### Change description ###

A Node process is single threaded by design. In order to leverage multicore VMs we have on our environments, [`cluster`](https://nodejs.org/api/cluster.html) module can be used to start... a cluster of Node processes.

```
[ ] Yes
[X] No
```


package.json
```diff
@@ -6,7 +6,7 @@
     "node": ">=8.0.0"
   },
   "scripts": {
-    "start": "./node_modules/.bin/ts-node -r tsconfig-paths/register src/main/server.ts",
```
rlewan: I'm bootstrapping paths programatically because the `-r` flag is not passed to forked processes. It probably can be done by settings `execArgv` on the `fork` call, but I thought that configuring this in code is cleaner in the end.

src/main/server.ts
```diff
@@ -1,26 +1,64 @@
...
+import * as cluster from 'cluster'
 
+const tsConfig = require('../../tsconfig.json')
+const numberOfCores = cpus().length
 const port: number = parseInt(process.env.PORT, 10) || 3000
-
 const logger = require('@hmcts/nodejs-logging').getLogger('server')
 
-if (app.locals.ENV === 'development' || app.locals.ENV === 'dockertests') {
+function bootstrapTypeScriptPaths (tsConfig: any) {
```
geek-soft: Would be great to have it (alongside `require('../../tsconfig.json')`) in separate file e.g. `ts-paths-bootstrap.ts`. Then we could do `import './ts-paths-bootstrap.ts'` at the beginning of the file.

The reason of the change is to separate concerns.

src/main/server.ts
```diff
@@ -1,26 +1,64 @@
...
+import * as tsConfigPaths from 'tsconfig-paths'
+import { cpus } from 'os'
+import * as express from 'express'
 import * as fs from 'fs'
 import * as path from 'path'
 import * as https from 'https'
+import * as cluster from 'cluster'
 
+const tsConfig = require('../../tsconfig.json')
+const numberOfCores = cpus().length
```
geek-soft: That should be more in local scope of `if (cluster.isMaster) {` block to keep declaration as close as possible to the place it is used.

rlewan:
Startup logging is now less verbose:
```
$ ./node_modules/.bin/ts-node src/main/server.ts
2017-10-30T16:56:45+01:00 INFO server: Testing support activated
2017-10-30T16:56:45+01:00 INFO server: Master process running on 60334
2017-10-30T16:56:48+01:00 INFO server: Listener started (PID 60337): https://localhost:3000
2017-10-30T16:56:48+01:00 INFO server: Listener started (PID 60335): https://localhost:3000
2017-10-30T16:56:48+01:00 INFO server: Listener started (PID 60338): https://localhost:3000
2017-10-30T16:56:48+01:00 INFO server: Listener started (PID 60336): https://localhost:3000
```

PR #157, opened 2017-10-30
ROC-2162: javascript version of evidence list page (progressive enhancement)

yarpo:
### JIRA link ###
https://tools.hmcts.net/jira/browse/ROC-2162

### Change description ###
Javascript progressive enhancement version

```
[ ] Yes
[x] No
```


timja-kainos:
javascript version of what?

yarpo:
@timja-kainos - javascript version of ROC-2162.

timja-kainos:
I more meant that the PR title wasn't the clearest, it assumes context

src/main/features/response/views/evidence.njk
```diff
@@ -22,20 +26,46 @@
     <select name="rows[{{ index }}][type]" id="rows[{{ index }}][type]"
       class="form-control-select {% if errorType %}form-control-error{% endif %}" >
 
-      <option value="">-- Please make a selection --</option>
+      <option value="">{{t('-- Please make a selection --') }}</option>
```
timja-kainos: can you add a space after `{{`

src/main/features/response/views/evidence.njk
```diff
@@ -1,6 +1,10 @@
 {% extends "layout.njk" %}
 {% from "form.njk" import csrfProtection, linkButton, textArea, saveAndContinueButton, errorSummary %}
```
timja-kainos: you removed textArea macro further down but not here?

src/main/features/response/views/evidence.njk
yarpo: done

src/main/features/response/views/evidence.njk
yarpo: done

src/main/features/response/views/evidence.njk
```diff
@@ -59,6 +89,7 @@
           {{ evidenceRow(loop.index0, form) }}
         {% endfor %}
         </div>
+        <br />
```
timja-kainos: do not use br, use css instead, prototype does this for hacking things together quickly

src/main/features/response/views/evidence.njk
```diff
@@ -22,20 +26,46 @@
...
-      }}
+    <div class="panel panel-border-narrow {% if not typeValue %} js-hidden {% endif %}">
+      {% set name = 'rows[' + index + '][description]' %}
+      {% set valueDescription = form.valueFor(name) %}
+      {% set errorDescription = form.errorFor(name) %}
+      <div class="form-group {% if errorDescription %} form-group-error {% endif %}">
+        <label for="{{ name }}" id="{{ name }}[label]" class="form-label">
+          {{ t('Describe this evidence in more detail (optional).') }}
+
+          <span class="evidence-message evidence-message-CONTRACTS_AND_AGREEMENTS {% if typeValue !== EvidenceType.CONTRACTS_AND_AGREEMENTS.value %}visually-hidden{% endif %}">
```
timja-kainos: should these be visually-hidden or js-hidden?

yarpo:
```
{% if typeValue !== EvidenceType.CONTRACTS_AND_AGREEMENTS.value %}visually-hidden{% endif %}
```

src/main/features/response/views/evidence.njk
```diff
@@ -22,20 +26,46 @@
...
+            {{ t('For example, a signed contract by the claimant.') }}
+          </span>
+          <span class="evidence-message evidence-message-EXPERT_WITNESS {% if typeValue !== EvidenceType.EXPERT_WITNESS.value %}visually-hidden{% endif %}">
```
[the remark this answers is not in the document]
geek-soft: Above will also remove `evidence-message-CONTRACTS_AND_AGREEMENTS`, `evidence-message-EXPERT_WITNESS` and others which are not being used.

src/main/features/response/views/evidence.njk
```diff
@@ -22,20 +26,46 @@
...
+          </span>
+          <span class="evidence-message evidence-message-RECEIPTS {% if typeValue !== EvidenceType.RECEIPTS.value %}visually-hidden{% endif %}">
+            {{ t('For example, a receipt showing you’ve paid the claimant.') }}
+          </span>
+          <span class="evidence-message evidence-message-STATEMENT_OF_ACCOUNT {% if typeValue !== EvidenceType.STATEMENT_OF_ACCOUNT.value %}visually-hidden{% endif %}">
+            {{ t('For example, a bank statement showing you’ve paid the claimant.') }}
+          </span>
+        </label>
+        {% if errorDescription %}
+          <span class="error-message">{{ t(error) }}</span>
```
geek-soft: I cannot see where `error` variable is defined :(

src/main/features/response/views/evidence.njk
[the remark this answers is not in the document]
yarpo: It would work well as evidence-message-CONTRACTS_AND_AGREEMENTS is needed to show a proper "for example" statement for specific option selected (javascript displays proper line and hides others).

src/main/features/response/views/evidence.njk
```diff
@@ -22,20 +26,46 @@
...
-            label = 'Describe this evidence in more detail (optional).',
-            name = 'rows[' + index + '][description]',
-            form = form,
-            inputClass='form-control timeline-column-text',
-            rows=2
-         )
-      }}
+    <div class="panel panel-border-narrow {% if not typeValue %} js-hidden {% endif %}">
+      {% set name = 'rows[' + index + '][description]' %}
+      {% set valueDescription = form.valueFor(name) %}
```
geek-soft: That and declaration below overrides declarations of the same name made few lines above. We should avoid overriding variables since that is error prone.

These are overridden:
```
{% set errorType = form.errorFor('rows[' + index + '][type]').length > 0 %}
{% set errorDescription = form.errorFor('rows[' + index + '][description]').length > 0 %}
```

Another thing is that from the names I would expect to have non boolean values while these are boolean. Probably we could name these better like `typeErrorExist` if we need these variables.

src/main/features/response/views/evidence.njk
```diff
@@ -1,5 +1,9 @@
 {% extends "layout.njk" %}
-{% from "form.njk" import csrfProtection, linkButton, textArea, saveAndContinueButton, errorSummary %}
+{% from "form.njk" import csrfProtection, linkButton, saveAndContinueButton, errorSummary %}
+
+{% block additional_scripts %}
+  <script src="{{ asset_paths['js'] }}/select-toggled-content.js"></script>
```
geek-soft: The way script is written makes it usable only for this particular use scenario. Name should reflect it so maybe we could have `envidence-select-handler.js` or similar.

src/main/public/js/select-toggled-content.js
```diff
@@ -0,0 +1,33 @@
...
+  $('.timeline-row .form-control-select').on('change', function () {
+
+    var $content = $(this).next()
+
+    if (this.value) {
+      $content
+        .removeClass('js-hidden')
+        .attr('aria-hidden', 'false')
+        .attr('aria-expanded', 'false')
+      $content
```
geek-soft: There is some repetition here, looks like text area is cleared on every change, no matter what value is selected

src/main/public/js/select-toggled-content.js
```diff
@@ -0,0 +1,33 @@
...
+        .find('.evidence-message-' + this.value)
+        .removeClass('visually-hidden')
+    } else {
+      $content
+        .addClass('js-hidden')
+        .attr('aria-hidden', 'true')
+        .attr('aria-expanded', 'false')
+      $content
+        .find('.evidence-message')
```
geek-soft: The same here, that could be extracted outside `if-else` statement.

src/main/public/js/select-toggled-content.js
```diff
@@ -0,0 +1,33 @@
+$(document).ready(function () {
+  $('.timeline-row .form-control-select').on('change', function () {
+
+    var $content = $(this).next()
+
+    if (this.value) {
+      $content
+        .removeClass('js-hidden')
+        .attr('aria-hidden', 'false')
+        .attr('aria-expanded', 'false')
```
geek-soft: I do not think `aria-expanded` should be here. This tag is used to indicate state of the element what expands content, not on the content being expanded itself.

src/main/public/js/select-toggled-content.js
```diff
@@ -0,0 +1,33 @@
+$(document).ready(function () {
+  $('.timeline-row .form-control-select').on('change', function () {
+
+    var $content = $(this).next()
+
+    if (this.value) {
+      $content
+        .removeClass('js-hidden')
+        .attr('aria-hidden', 'false')
```
geek-soft: A `js-hidden` hides element for both screenreaders and browsers so I do not think `aria-hidden` applies here. If something is removed from DOM it won't be visible to screen readers.

src/main/public/js/select-toggled-content.js
yarpo: done

src/main/public/js/select-toggled-content.js
yarpo: done

src/main/features/response/views/evidence.njk
```diff
@@ -1,41 +1,70 @@
 {% extends "layout.njk" %}
-{% from "form.njk" import csrfProtection, linkButton, textArea, saveAndContinueButton, errorSummary %}
+{% from "form.njk" import csrfProtection, linkButton, saveAndContinueButton, errorSummary %}
+
+{% block additional_scripts %}
+  <script src="{{ asset_paths['js'] }}/envidence-select-handler.js"></script>
```
timja-kainos: typo: envidence  -> evidence

src/main/features/response/views/evidence.njk
yarpo: renamed

PR #161, opened 2017-10-31
Added influxdb step to pipeline

jrjohnson87:
**Before creating a pull request make sure that:**

- [ ] commit messages are meaningful and follow good commit message guidelines
- [ ] README and other documentation has been updated / added (if needed)
- [ ] tests have been updated / new tests has been added (if needed)

Please remove this line and everything above and fill the following sections:

### Change description ###

```
[ ] Yes
[ ] No
```


Jenkinsfile
```diff
@@ -164,4 +164,16 @@ timestamps {
     milestone()
   }
   notifyBuildFixed channel: channel
+
+  if (exit_status != 0) {
```
timja-kainos: as far as I know this shouldn't be needed?
those vars should be set already?
 
I think you want these in  a `finally` block so that they always execute

jrjohnson87: Honestly not sure, just going on the docs:

> Pipelines don't have post-build actions, so the build result, build ordinal, and the build success boolean must be set manually. They will default to "?", 5, and false respectively unless set before calling the step. The boolean value is set according to the ordinal value and the ordinal value is fetched according to build result, so only the build result have to be set manually. Also, the build status will appear as "?" and the build duration might be a little off, because the build is not actually finished. If you want to get those pieces of information you need to configure the plugin separately on each job as a post-build action. The jobs can be run with, for example, the Build Pipeline Plugin to get data from all jobs to InfluxDB. Alternatively, you can insert the information in your build manually inside your Groovy script.

See https://wiki.jenkins.io/display/JENKINS/InfluxDB+Plugin

timja-kainos: hmm, the docs suggest setting it to success as the last thing in the try block, setting it to failure in the catch block and then in a finally block do the step.
make sense?

timja-kainos:
Closing as we aren't using the jenkins step any more
See https://github.com/hmcts/cmc-citizen-frontend/pull/210

PR #162, opened 2017-10-31
Update UI to make it more consistent

tarik-johnston1:
Changed the page title x-large font to large
Added two thirds to check and send & response type pages
Changed colour of report a problem css and increased padding
Added class name of form group to details element for increased padding


src/main/public/stylesheets/components/_report_problem.scss
```diff
@@ -12,9 +13,14 @@ $report-problem-colour: #454a4c;
...
   font-size: 16px;
   margin: 90px auto -80px 0;
+  padding: 50px 0 0;
 
   span {
     display: block;
     font-size: 14px;
   }
+
+  a {
```
timja-kainos: won't this change all links to report-problem-colour? Is that intended?

geek-soft: Should be fine as `a` is inside `.report-error` block. It will affect all elements matching `.report-error a`.

geek-soft:
@timja-kainos, coverage is funny here, few content links updated and coverage increased by +0.6% :(

src/main/public/stylesheets/components/_report_problem.scss
timja-kainos: aah right, didn't see it, had to open above the diff to see it 🙍

timja-kainos:
Should pass after this builds: https://github.com/hmcts/cmc-integration-tests/pull/18

PR #163, opened 2017-10-31
Remove unnecessary translations before passing text arguments to macros

robertparkinson:


src/main/features/claim/views/interest-date.njk
```diff
@@ -38,7 +38,7 @@
               {{ textInput('Explain why you’re claiming from this date', 'reason', form) }}
 
             </div>
-            {{ radioOption(t("Claim interest from the date I submit the claim"), 'type', form, 'submission', ' ', form.valueFor('type') == 'submission') }}
```
yarpo: can you please also replace `"` with `'`?

package.json
```diff
@@ -21,11 +21,11 @@
     "prepush": "yarn test && yarn test:routes"
   },
   "dependencies": {
-    "@hmcts/cmc-common-frontend": "^1.10.3",
+    "@hmcts/cmc-common-frontend": "^1.12.1",
```
geek-soft: There is already v 1.13.0 released.

PR #168, opened 2017-11-02
ROC-2253: Use common macro for displaying claim status

yarpo:
### JIRA link ###
https://tools.hmcts.net/jira/browse/ROC-2253

### Change description ###
Using macros from common-fe project

```
[ ] Yes
[x] No
```


src/test/features/response/form/models/responseType.ts
```diff
@@ -0,0 +1,28 @@
+import { expect } from 'chai'
+import { ResponseType } from 'response/form/models/responseType'
+
+describe('ResponseType', () => {
+
+  describe('all', () => {
+
+    it('should return array of 3', () => {
```
timja-kainos: this test doesn't add value imo, it means you have to change it if you add a response type but why does the test care?

src/test/features/response/form/models/responseType.ts
```diff
@@ -0,0 +1,28 @@
...
+      const actual: ResponseType[] = ResponseType.all()
+      expect(actual.length).to.be.eq(3)
+    })
+
+  })
+
+  describe('valueOf', () => {
+
+    it('should return ResponseType', () => {
+      expect(ResponseType.valueOf(ResponseType.OWE_ALL_PAID_NONE.value)).to.be.eq(ResponseType.OWE_ALL_PAID_NONE)
```
timja-kainos: not sure of the value of this test, especially how its checking all options

src/main/features/dashboard/views/defendant.njk
```diff
@@ -20,13 +22,13 @@
         <p>{{ t('If you don’t owe the money you can <a href="%s" target="blank">ask the court to cancel</a> the judgment.',
               { postProcess: 'sprintf', sprintf: [ 'https://www.gov.uk/county-court-judgments-ccj-for-debt/if-you-dont-owe-the-money' ]}) | safe }}</p>
       {% else %}
-        <p>{{ t('Your defence is being processed. We will email you to explain what to do next.') }}</p>
```
geek-soft: @yarpo , is that removal part of the story?

src/test/features/response/form/models/responseType.ts
```diff
@@ -0,0 +1,28 @@
...
+      expect(ResponseType.valueOf(ResponseType.OWE_NONE.value)).to.be.eq(ResponseType.OWE_NONE)
+      expect(ResponseType.valueOf(ResponseType.OWE_SOME_PAID_NONE.value)).to.be.eq(ResponseType.OWE_SOME_PAID_NONE)
+    })
+
+    it('should return undefined', () => {
```
rlewan: `should return undefined for incorrect value`

src/test/features/response/form/models/responseType.ts
```diff
@@ -0,0 +1,28 @@
...
+    it('should return array of 3', () => {
+      const actual: ResponseType[] = ResponseType.all()
+      expect(actual.length).to.be.eq(3)
+    })
+
+  })
+
+  describe('valueOf', () => {
+
+    it('should return ResponseType', () => {
```
rlewan: `should return ResponseType for corresponding value`

src/main/features/dashboard/views/defendant.njk
yarpo: According to Julian we should display the same status on the main dashboard and here. This was very generic message that was replaced with the proper one.

src/test/features/response/form/models/responseType.ts
yarpo: ok

src/main/features/dashboard/views/index.njk
```diff
@@ -44,25 +44,13 @@
             </tr>
           {% endif %}
           {% for claim in claimsAsClaimant %}
-            {% if claim.eligibleForCCJ %}
```
geek-soft: Something to improve in the future. Evaluating status in template is a bad idea because it is difficult to test. We should have class evaluating status, well tested. Then we pass status to the template to match it with the text for end user.

src/test/features/response/form/models/responseType.ts
geek-soft: Yeah, that does not protect use from removing one of existing states and replacing it with new one.

src/main/features/dashboard/views/defendant.njk
geek-soft: Thanks for explaining this to me.

src/test/features/dashboard/routes/claimant.ts
```diff
@@ -50,7 +50,7 @@ describe('Dashboard - claimant page', () => {
           await request(app)
             .get(claimantPage)
             .set('Cookie', `${cookieName}=ABC`)
-            .expect(res => expect(res).to.be.successful.withText('Claim number', 'The defendant'))
+            .expect(res => expect(res).to.be.successful.withText('Claim number', 'Request a County Court Judgment'))
```
geek-soft: Maybe just `withText('Claim number: ' + claimNumber)` would be enough here. A `Request a County Court Judgment` is very much depends on the state of the claim

src/test/features/dashboard/routes/defendant.ts
```diff
@@ -46,7 +46,7 @@ describe('Dashboard - defendant page', () => {
           await request(app)
             .get(defendantPage)
             .set('Cookie', `${cookieName}=ABC`)
-            .expect(res => expect(res).to.be.successful.withText('Claim number', 'Your defence is being processed'))
+            .expect(res => expect(res).to.be.successful.withText('Claim number', 'Respond to claim'))
```
geek-soft: The same here, just looking form simplification.

src/test/features/response/form/models/responseType.ts
geek-soft: If we wanna have this text we could have a loop over result of `all` and then make sure `ResponseType.valueOf` can return valid value for every possible 'instance' of ResponseType. Current implementation will give us more maintenance effort.

src/main/features/dashboard/views/index.njk
yarpo: I'd leave it as it is now, as this story is taking too long.

This story also doesn't cover all possible statuses as we haven't implement all paths yet (hands off). Once all paths are implemented and we have very final version of statuses and dashboards I'd be more than happy to move it to a proper unit testable class.

src/test/features/dashboard/routes/claimant.ts
yarpo: OK, I'll change it. I just fixed the test that was broken when we started displaying more specific status. The previous one was VERY generic ;)

src/test/features/dashboard/routes/defendant.ts
yarpo: ok

src/test/features/dashboard/routes/claimant.ts
yarpo: done

src/test/features/dashboard/routes/defendant.ts
yarpo: done

src/main/features/dashboard/views/index.njk
```diff
@@ -44,25 +44,13 @@
...
-
             <tr class="claims default mobile-table">
               <td><a href="/dashboard/{{ claim.externalId }}/claimant">{{ claim.claimNumber }}</a></td>
               <td>{{ claim.claimData.defendant.name }}</td>
               <td>{{ claim.totalAmount | numeral }}</td>
-              <td class="mobile-hide">{{ status }}</td>
+              <td class="mobile-hide">{{ responseStatusForClaimant(claim) }}</td>
             </tr>
-            <tr><td colspan="3" class="mobile-table-status">{{ t('Status: ') + status }}</td></tr>
+            <tr><td colspan="3" class="mobile-table-status">{{ t('Status: ') + responseStatusForClaimant(claim) }}</td></tr>
```
comanirudha: may be {{ t('Status: %s', { postProcess: 'sprintf', sprintf: [ responseStatusForClaimant(claim) ]})}}

src/main/features/dashboard/views/index.njk
geek-soft: Definitely not for now, next time possibly when we touch statuses.

src/main/features/dashboard/views/index.njk
geek-soft: Not sure that would work, `responseStatusForClaimant ` does not return anything, just prints HTML markup. Worth trying, was curious myself if that would work.

src/test/features/response/form/models/responseType.ts
```diff
@@ -0,0 +1,19 @@
+import { expect } from 'chai'
+import { ResponseType } from 'response/form/models/responseType'
+
+describe('ResponseType', () => {
+
+  describe('valueOf', () => {
+
+    it('should return ResponseType for corresponding value', () => {
+      expect(ResponseType.valueOf(ResponseType.OWE_ALL_PAID_NONE.value)).to.be.eq(ResponseType.OWE_ALL_PAID_NONE)
```
geek-soft: Shall we make it more dynamic by iterating through all types?

timja-kainos:
Conflicts @yarpo

src/test/features/response/form/models/responseType.ts
yarpo: I don't mind. @rlewan, @timja-kainos what do you think? Do you like this idea?

rlewan: I'm already doing it for the residence question
```
    it('should return correct values', () => {
      ResidenceType.all().forEach((element) => {
        expect(ResidenceType.valueOf(element.value).value).to.equal(element.value)
      })
    })
```
so yes -- I like the idea ;)

timja-kainos: sure

PR #170, opened 2017-11-02
chore(package): update dependencies

KarolJastrzebski:
### Change description ###

Enable greenkeeper

```
[ ] Yes
[x] No
```


package.json
```diff
@@ -8,7 +8,7 @@
   "scripts": {
     "start": "./node_modules/.bin/ts-node -r tsconfig-paths/register src/main/server.ts",
     "setup": "gulp sass copy-files",
-    "lint": "tslint --project tsconfig.json --type-check && sass-lint -v -q",
```
geek-soft: @timja-kainos why we are removing this?

timja-kainos: have you not noticed the spam that's been happening everytime you push for the last week 😉 
type-check is deprecated doing --project implies it or something

geek-soft: Alright, deprecated in 5.8 👍 

> [deprecation] CLI argument --type-check is no longer necessary and will be removed in the next major version (#3322)

PR #172, opened 2017-11-03
ROC-2431: Compile TS files in Dockerfile

geek-soft:
### JIRA link ###

https://tools.hmcts.net/jira/browse/ROC-2431

### Change description ###

With this change when docker image is build `tsc` compiles TS files in JS so that docker process just runs JS files. That speeds up application startup time in the container especially when container is started for the first time when TS cache is empty.

Full compilation of the TS files with `ts-node` for this application takes up to 60 seconds. The same with `tsc` takes just around 7 seconds.

In the future we might consider removing ts-node in favour of webpack or regular TS watchers.

```
[ ] Yes
[x] No
```


src/main/common/router/routerFinder.ts
```diff
@@ -1,8 +1,11 @@
+import * as path from 'path'
 import { Router } from 'express'
 import * as requireDirectory from 'require-directory'
 
+const fileExtension: string = path.extname(__filename).slice(1)
```
geek-soft: If `ts-node` process runs an application the extension is equal `ts`. If node is used then extension is obviously `js`.

package.json
```diff
@@ -6,8 +6,10 @@
     "node": ">=8.0.0"
   },
   "scripts": {
-    "start": "./node_modules/.bin/ts-node -r tsconfig-paths/register src/main/server.ts",
+    "compile": "tsc --project tsconfig.prod.json",
```
timja-kainos: I think this should be run as part of setup so that it runs on the environments

package.json
```diff
@@ -6,8 +6,10 @@
...
     "setup": "gulp sass copy-files",
+    "start": "./node_modules/.bin/ts-node -r tsconfig-paths/register src/main/server.ts",
```
timja-kainos: I think this should be inverted, to start and start-dev, so that the environments startup quicker as well

src/main/common/router/routerFinder.ts
```diff
@@ -1,8 +1,11 @@
...
+
 const options: object = {
-  extensions: ['ts'],
+  extensions: [fileExtension],
```
timja-kainos: could just change this to accept js again, I only removed it last week

package.json
geek-soft: I think you are right, I had the same thought. Then possibly just gulp would use `ts-dev`.

src/main/common/router/routerFinder.ts
geek-soft: Did that as a first thing but faced some issue. Because JS files in the same place are TS files (right next to TS files) it `routeFinder` loaded both and then there was compilation error because `ts-node` was unable to work with compiled JS files.

Dockerfile
```diff
@@ -1,16 +1,18 @@
-FROM node:8.1.4
+FROM node:8.9.0-alpine
 
 RUN mkdir -p /usr/src/app
 WORKDIR /usr/src/app
 
-COPY package.json yarn.lock /usr/src/app/
-RUN yarn install
+COPY package.json yarn.lock tsconfig.json tsconfig.prod.json gulpfile.js /usr/src/app/
+RUN yarn install \
```
geek-soft: Installing prod dependencies only and clearing cache reduces image size from around 1.1GB to 768MB. Swapping to alpine image reduces the size further to 161MB :)

timja-kainos: futher 161MB or to 161MB image size?

geek-soft: To 161MB, that will be now the image size.

geek-soft: By the way build sequence might be optimised a bit more, will do that on Monday.

Dockerfile
```diff
@@ -1,16 +1,21 @@
-FROM node:8.1.4
+FROM node:8.9.0-alpine
 
 RUN mkdir -p /usr/src/app
 WORKDIR /usr/src/app
 
 COPY package.json yarn.lock /usr/src/app/
```
geek-soft: I left this a separate step to speed up build time. If copying source files happened before `yarn install` it would mean that every tiniest file change would trigger full dependencies installation.

Unfortunately that increases image size to around ~500MB because yarn cache remains as separate layer.

package.json
geek-soft: Done. Next step is to remove `ts-node` completely. I did not want to do it because it would require slightly more work to remove compilation from test scripts.

PR #177, opened 2017-11-07
Remove and ban default exports

geek-soft:
### JIRA link ###

None

### Change description ###

This PR does:

1. remove default exports from existing code base
2. forbids using default exports through tslint

Unfortunately it is required to disable lint rule in route files since they use default export by default. Exporting constants looked like not an good option so I decided to keep default exports there for now.

```
[ ] Yes
[x] No
```


src/main/features/ccj/routes/pay-by-set-date.ts
```diff
@@ -4,14 +4,15 @@ import { Paths } from 'ccj/paths'
...
-import User from 'idam/user'
+import { User } from 'idam/user'
 import { DraftService } from 'services/draftService'
 import { ErrorHandling } from 'common/errorHandling'
 
 function renderView (form: Form<PayBySetDate>, res: express.Response): void {
   res.render(Paths.payBySetDatePage.associatedView, { form: form })
 }
 
+/* tslint:disable:no-default-export */
```
timja-kainos: any plan on a way to address this?

geek-soft: Not at this moment. I would need to think how to rewrite routes so that we can have a nice code without default exports.

PR #189, opened 2017-11-09
ROC-2476: calling api to reject or accept an offer

yarpo:
### JIRA link ###
https://tools.hmcts.net/jira/browse/ROC-2476

### Change description ###
Calling API endpoints to store informations about 

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/app/claims/claimStoreClient.ts
+++ b/src/main/app/claims/claimStoreClient.ts
@@ -8,6 +8,7 @@ import { OfferModelConverter } from 'claims/offerModelConvertor'
 import { Offer } from 'claims/models/offer'
 import { Offer as OfferForm } from 'features/offer/form/models/offer'
 import { ForbiddenError } from '../../errors'
+
 export const claimApiBaseUrl: string = `${config.get<string>('claim-store.url')}`
 export const claimStoreApiUrl: string = `${claimApiBaseUrl}/claims`
 const claimStoreResponsesApiUrl: string = `${claimApiBaseUrl}/responses/claim`
@@ -35,6 +36,7 @@ export class ClaimStoreClient {
       }
     })
   }
+
   static saveOfferForUser (madeBy: string, user: User, offerForm: OfferForm): Promise<void> {
     const claim: Claim = user.claim
     const offer: Offer = OfferModelConverter.convert(offerForm)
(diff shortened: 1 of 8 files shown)


src/main/app/claims/claimStoreClient.ts
```diff
@@ -45,6 +47,27 @@ export class ClaimStoreClient {
       }
     })
   }
+
+  static acceptOfferByUser (acceptedBy: string, user: User): Promise<Claim> {
```
rlewan: Would be nice to rename methods in this class to have just `makeOffer`, acceptOffer` and `rejectOffer`.

rlewan: What do you think about having a separate client for it as well?

src/test/http-mocks/claim-store.ts
```diff
@@ -236,11 +241,25 @@ export function rejectSaveOfferForDefendant (reason: string = 'HTTP error') {
...
 }
+
 export function resolveSaveOffer () {
   mock(`${serviceBaseURL}/claims`)
     .post(new RegExp('/[0-9]+/offers/defendant'))
-    .reply(HttpStatus.OK)
+    .reply(HttpStatus.CREATED)
 }
+
+export function resolveAcceptOffer (by: string = 'claimant') {
```
rlewan: What do you think about testing for failures as well?

geek-soft: Should't that be in capitals? Would be possible to reuse `MadeBy` buy repurposing the class so we have `OfferParticipantType.CLAIMANT.value`. I am pretty sure we could find better name as well :)

src/main/app/claims/claimStoreClient.ts
yarpo: I was thinking about it as we have a separate client for CCJ as well.

src/test/http-mocks/claim-store.ts
yarpo: It doesn't matter if it's upper or lower case. Both works for the same endpoint.

This is actually weird we don't have consts like this so far on frontend. If we want to introduce it what do you think about MadeBy enum (as we have one in API)? The name is more generic and we can use it in more contexts.

PR #191, opened 2017-11-10
ROC-2556: Question 1 - Where do you live?

rlewan:
### JIRA link ###

https://tools.hmcts.net/jira/browse/ROC-2556

### To Do

- [x] Update check and send page
- [x] update `WhenWillYouPayTask.isCompleted` to cover SOM
- [x] Make sure pages look good on mobile
- [ ] Align integration tests

```
[ ] Yes
[X] No
```


src/main/features/response/views/statement-of-means/residence.njk
```diff
@@ -0,0 +1,76 @@
...
+        data-event-label-from="type[value]"
+        novalidate>
+
+        {{ csrfProtection(csrf) }}
+
+        <div class="form-group {% if error %} form-group-error {% endif %}">
+          <fieldset id="type[label]">
+            <legend class="visually-hidden">{{ t('Where are you living?') }}</legend>
+
+            {{ radioOption(
```
geek-soft: It looks like it should be possible to use `radioGroup` macro here.

.gitignore
```diff
@@ -1,4 +1,3 @@
-src/**/*.js
```
timja-kainos: what for?

rlewan: I've accidentally compiled TS to JS in IDEA and spent 30 minutes working out why my code crashes and I get stack traces of JS files. This `.gitignore` entry hid the problem and should be removed in my opinion. Doesn't give any value really.

rlewan: Context: given an option `ts-node` decided to load JS files over TS.

timja-kainos: I hit the same problem this morning, and lost a similar amount of time, I ran `yarn compile` to reproduce an issue jenkins was having.

hmm, sounds reasonable. @geek-soft any objection?

src/main/features/response/form/models/statement-of-means/residence.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  static fromObject (input?: any) {
+    if (!input) {
+      return input
+    }
+    const type: ResidenceType = ResidenceType.valueOf(input.type)
+    return new Residence(
+      type,
+      (type && type.value === ResidenceType.OTHER.value) ? input.housingDetails : undefined
```
timja-kainos: is this `type &&` check needed?

.gitignore
geek-soft: Not really, we can try it however real issue is that `ts-node` loads JS files and we should solve that first.

src/main/features/response/form/models/statement-of-means/residence.ts
geek-soft: Would `type === ResidenceType.OTHER` work as well? A `ResidenceType` is a singleton in Node env so that should work.

src/main/features/response/views/statement-of-means/residence.njk
```diff
@@ -0,0 +1,55 @@
...
+
+{% block content %}
+  <div class="grid-row">
+    <div class="column-two-thirds">
+      <h1 class="heading-large">{{ t('Where are you living?') }}</h1>
+
+      <form
+        method="POST"
+        class="analytics-submit-event-trigger"
+        data-event-action="Response: Statement of means: Select defendant response option"
```
geek-soft: An action does not sound good here. We asking for `Where are you living?` while we sending an action to GA as `Select defendant response option`

src/main/features/response/views/statement-of-means/residence.njk
```diff
@@ -0,0 +1,55 @@
...
+
+        {{ csrfProtection(csrf) }}
+
+        {% set error = form.errorFor('type') %}
+        <div class="form-group {% if error %} form-group-error {% endif %}">
+          <fieldset id="type[label]">
+            <legend class="visually-hidden">{{ t('Where are you living?') }}</legend>
+            {% if error %}<span class="error-message">{{ t(error) }}</span>{% endif %}
+
+            {% for type in ResidenceType.all() %}
```
geek-soft: I would suggest implementing `ResidenceType#except` method as we tend to do. Then we could do `ResidenceType.except(ResidenceType.OTHER)` in the view. Alternatively we could do that in the route file.

src/test/features/response/draft/responseDraft.ts
```diff
@@ -32,6 +33,7 @@ describe('ResponseDraft', () => {
       expect(responseDraftModel.moreTimeNeeded.option).to.eql(MoreTimeNeededOption.YES)
       expect(responseDraftModel.isMoreTimeRequested()).to.be.eql(true)
       expect(responseDraftModel.impactOfDispute.text).to.equal('This dispute has affected me badly, I cried')
+      expect(responseDraftModel.statementOfMeans.residence.type.value).to.equal(ResidenceType.OTHER.value)
```
geek-soft: Would `expect(responseDraftModel.statementOfMeans.residence.type).to.equal(ResidenceType.OTHER)` work?

src/test/features/response/draft/statementOfMeans.ts
```diff
@@ -1,14 +1,23 @@
...
 import { expect } from 'chai'
 import { StatementOfMeans } from 'response/draft/statementOfMeans'
 import { Residence } from 'response/form/models/statement-of-means/residence'
 import { Employment } from 'response/form/models/statement-of-means/employment'
 import { Employers } from 'response/form/models/statement-of-means/employers'
 import { SelfEmployed } from 'response/form/models/statement-of-means/selfEmployed'
+import { ResidenceType } from 'response/form/models/statement-of-means/residenceType'
 
 describe('StatementOfMeans', () => {
+  context('when creating a new instance', () => {
```
geek-soft: I recently had a change of heart in terms of conductor test to the level I would not do it anymore. Constructors are not something we see being tested too ofter I think.

.gitignore
radoslawn-kainos: This thing happened to me once too.

@geek-soft 
I agree that ts-node should give preference to TS files over JS, but I think at least part of the problem is that build output files are put in src directory. If we maintain good configuration hygiene and keep our input and output separate, we won't need to depend on the ts-node behaviour as much. Also, all we would need in .gitignore to handle this is a single entry for the build output directory.

PR #192, opened 2017-11-10
Hide offer links when feature is disabled

geek-soft:
### JIRA link ###

Covers many stories

### Change description ###

I checked all places where we link to offer pages and wrapped these made from outside of the offer feature with feature toggle.

I also registered globally OfferPaths and used these in all templates. I think doing this in the templates by default has an advantage of having single place to check when we look for links in the templates. Also fact that feature registers nunchucks variable itself is nice as well. I am happy to apply this pattern to remaining parts of the application if it gets good reception.

At the very end I pre-converted all feature toggles to boolean type so no need to do it in the templates. That surprised me a bit when I realised type is not a boolean so to avoid errors I made conversion at a registration time.

```
[ ] Yes
[x] No
```


src/main/features/dashboard/routes/index.ts
```diff
@@ -20,7 +19,6 @@ export default express.Router()
 
     res.render(Paths.dashboardPage.associatedView, {
       paths: Paths,
-      offerPaths: OfferPaths,
```
geek-soft: Instead of having many different names global variable enforces single one.

src/main/features/offer/guards/offerGuard.ts
```diff
@@ -1,24 +0,0 @@
...
-import { Claim } from 'claims/models/claim'
-import { Paths } from 'response/paths'
-import * as config from 'config'
-import * as toBoolean from 'to-boolean'
-import { User } from 'idam/user'
-
-const logger = require('@hmcts/nodejs-logging').getLogger('response/guards/offerGuard')
-const eligibleForOffer = toBoolean(config.get<boolean>('featureToggles.offer'))
-
-export class OfferGuard {
```
geek-soft: No need for such guard since feature is toggled itself.

src/main/features/offer/index.ts
```diff
@@ -20,6 +21,10 @@ function requestHandler (): express.RequestHandler {
 
 export class Feature {
   enableFor (app: express.Express) {
+    if (app.settings.nunjucksEnv && app.settings.nunjucksEnv.globals) {
```
geek-soft: This pattern allows modules to register variable with Nunjucks. We can create base class which makes it slightly more convenient and apply in parts of the app. Then Nunjucks module does not have to be overload with registering all elements of the app.

src/main/features/offer/routes/response.ts
```diff
@@ -51,9 +48,8 @@ export default express.Router()
             await OfferClient.rejectOffer(user)
             res.redirect(Paths.rejectedPage.evaluateUri({ externalId: user.claim.externalId }))
             break
-
           default:
-            res.redirect(Paths.responsePage.evaluateUri({ externalId: user.claim.externalId }))
+            throw new Error(`Option ${form.model.option} is not supported`)
```
geek-soft: I prefer hard error instead of 'refresh' redirect which makes it easier to spot problems. When counter offer is implemented it will go away.

PR #199, opened 2017-11-14
ROC-2820: Refactor Jenkinsfile

timja-kainos:
Explicitly do onPR and onMaster sections to be clear what runs where

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -17,7 +17,7 @@ Packager packager = new Packager(this, 'cmc')
 
 SmokeTests smokeTests = new SmokeTests(this)
 IntegrationTests integrationTests = new IntegrationTests(env, this)
-def channel = '#cmc-tech-notification'
+String channel = '#cmc-tech-notification'
 
 timestamps {
   milestone()
@@ -34,28 +34,15 @@ timestamps {
           checkout scm
         }
 
-        stage('Setup') {
-          sh '''
-            yarn install
-            yarn setup
-          '''
-        }
-
-        onPR {
-          stage('Package application (Docker)') {
-            citizenFrontendVersion = dockerImage imageName: 'cmc/citizen-frontend'
-          }
-
-          stage('Integration Tests') {
-            integrationTests.execute([
-              'CITIZEN_FRONTEND_VERSION': citizenFrontendVersion,
-              'TESTS_TAG'               : '@citizen'
-            ])
-          }
-        }
-
         // Travis runs everything except integration tests, no need to do this twice (but run on master to be safe)
         onMaster {
+          stage('Setup') {
+            sh '''
+              yarn install
+              yarn setup
+            '''
+          }
+
           stage('Node security check') {
             try {
               sh "yarn test:nsp 2> nsp-report.txt"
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -90,50 +103,24 @@ timestamps {
...
-        }
-
-        stage('Package application (Docker)') {
-          citizenFrontendVersion = dockerImage imageName: 'cmc/citizen-frontend'
-        }
 
-        stage('Package application (RPM)') {
-          citizenFrontendRPMVersion = packager.nodeRPM('citizen-frontend')
-          version = "{citizen_frontend_buildnumber: ${citizenFrontendRPMVersion}}"
+          stage('Package application (RPM)') {
```
geek-soft: @rlewan, @timja-kainos , @yarpo what your opinion on building RPM's on PR branch as well?

Jenkinsfile
```diff
@@ -41,20 +41,33 @@ timestamps {
           '''
```
geek-soft: I have a feeling there is no point in running `Setup` stage with `yarn install` and `yarn setup` on PR since docker packages does not need it.

Jenkinsfile
rlewan: How likely is it that RPM build will fail? Not too likely in my opinion and a possibility of master build failing because of that is a risk I'm willing to take 👍

Jenkinsfile
```diff
@@ -90,50 +103,24 @@ timestamps {
               archiveArtifacts 'coverage-report/lcov-report/index.html'
             }
           }
-        }
-
-        stage('Sonar') {
-          onPR {
```
geek-soft: We are loosing this, anyone to miss it? There was no much value in this so far I we couldn't get sonar plugin to comment on PR's so we were unable to really see any issue.

Jenkinsfile
```diff
@@ -90,50 +103,24 @@ timestamps {
...
-            """
-          }
-
-          onMaster {
+          stage('Sonar') {
             sh "yarn sonar-scanner -- -Dsonar.host.url=$SONARQUBE_URL"
           }
-        }
-
-        stage('Package application (Docker)') {
```
geek-soft: I think we should still build Docker images and run integration tests on master.

This is also needed (by default) to tag docker images with latest tag so that we can pull these from Artifactory locally.

Jenkinsfile
geek-soft: Agree with you @rlewan. I am a bit in between but more keen to do it only on master.

Jenkinsfile
rlewan: Agree, they should be run on master as well.

Jenkinsfile
rlewan: If we see this is causing trouble we can always bring building RPMs on PR back.

geek-soft:
That should make a PR pipeline faster by circa 90 seconds.

Jenkinsfile
timja-kainos: In the many months doing this I've only seen it break twice for legitimate reasons, and a few times for jenkins being full of disk space. Not worth doing it on PRs imo

Jenkinsfile
timja-kainos: right, oops

PR #202, opened 2017-11-15
ROC-2558: VERY basic structure of folders for new feature - Statement of Means

yarpo:
### JIRA link ###
https://tools.hmcts.net/jira/browse/ROC-2558

### Change description ###
There are 3 stories about Statement of means being played simultaneously at the moment and we don't have a proper feature-like structure for it.

I decided to add it just to make sure we will not have too many conflicts. Everything is hidden by feature toggle flag so it's not breaking anything.  

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/features/statement-of-means/paths.ts
+++ b/src/main/features/statement-of-means/paths.ts
@@ -1,0 +1,7 @@
+import { RoutablePath } from 'common/router/routablePath'
+
+const statementOfMeansPath: string = '/case/:externalId/statement-of-means'
+
+export class Paths {
+  static readonly employmentPage = new RoutablePath(`${statementOfMeansPath}/employment`)
+}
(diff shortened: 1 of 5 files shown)


src/main/features/statement-of-means/paths.ts
```diff
@@ -0,0 +1,7 @@
+import { RoutablePath } from 'common/router/routablePath'
+
+const statementOfMeansPath: string = '/case/:externalId/statement-of-means'
+
+export class Paths {
+  static readonly employmentPage = new RoutablePath(`${statementOfMeansPath}/employment`)
```
timja-kainos: we already had this conversation on Rafal's PR for setting this up :(, statement of means is under response which is why it doesn't have a feature setup

yarpo:
OK, it's going to be a part of response flow. I have bad feelings about the size of response feature in the future. But let's keep it consistent.

PR #211, opened 2017-11-17
Fix total claim amount on PDF issue receipts

rlewan:
### Change description ###

Total claim amount was missing on issue receipts (both claim and response PDFs).

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/app/pdf/issueReceipt.ts
+++ b/src/main/app/pdf/issueReceipt.ts
@@ -4,6 +4,7 @@ import { Claim } from 'claims/models/claim'
 import { PartyDetailsMapper } from 'app/pdf/mappers/partyDetailsMapper'
 import { TheirDetailsMapper } from 'app/pdf/mappers/theirDetailsMapper'
 import { ClaimMapper } from 'app/pdf/mappers/claimMapper'
+import { NumberFormatter } from 'utils/numberFormatter'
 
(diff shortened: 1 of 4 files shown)


src/main/app/pdf/issueReceipt.ts
```diff
@@ -19,6 +20,7 @@ export class IssueReceipt {
   data (): object {
     return {
       claim: ClaimMapper.createClaimDetails(this.claim),
+      totalClaimAmount: NumberFormatter.formatMoney(this.claim.totalAmountTillToday),
```
timja-kainos: I think this should be: `totalAmountTillDateOfIssue` @rlewan  @pragnesh-patel

src/main/app/pdf/responseReceipt.ts
```diff
@@ -20,6 +21,7 @@ export class ResponseReceipt {
   data (): object {
     const data = {
       claim: ClaimMapper.createClaimDetails(this.claim),
+      totalClaimAmount: NumberFormatter.formatMoney(this.claim.totalAmountTillToday),
```
timja-kainos: same here

src/main/app/pdf/issueReceipt.ts
rlewan: Yeah, you're right 🙈

PR #216, opened 2017-11-22
Add missing type information for instances in local response scope

geek-soft:
### JIRA link ###

None

### Change description ###

This PR extracts variables out of the `res.locals.user...`. Having type definitions in place should help us avoid issues with refactorings that break the application. 

```
[ ] Yes
[x] No
```


src/main/features/response/routes/response-type.ts
```diff
@@ -19,7 +20,9 @@ function renderView (form: Form<Response>, res: express.Response) {
 export default express.Router()
   .get(Paths.responseTypePage.uri,
     ErrorHandling.apply(async (req: express.Request, res: express.Response, next: express.NextFunction) => {
-      renderView(new Form(res.locals.user.responseDraft.document.response), res)
+      const draft: ResponseDraft = res.locals.user.responseDraft.document
```
timja-kainos: no user being extracted here?

src/main/features/response/routes/more-time-request.ts
```diff
@@ -27,7 +28,9 @@ export default express.Router()
     Paths.moreTimeRequestPage.uri,
     MoreTimeAlreadyRequestedGuard.requestHandler,
     async (req: express.Request, res: express.Response, next: express.NextFunction) => {
-      renderView(new Form(res.locals.user.responseDraft.document.moreTimeNeeded), res, next)
+      const draft: ResponseDraft = res.locals.user.responseDraft.document
```
timja-kainos: same here

src/main/features/response/routes/response-type.ts
geek-soft: I did it on purpose since sometimes a user variable is not needed directly because we just need a claim or one of the drafts. I decided not to extract user variable in such cases as my thinking was that we still get type information needed in given scope.

I am aware that it might still make us some issue when we refactor User or Draft classes but I was happy to take the risk to make code a bit nicer. I knew it might trigger questions so I am open for conversation

Note: Also in the future I am thinking that claim or draft properties should be stored in `res.locals` map directly not as a sub property of user instance.

timja-kainos: I would prefer if we were consistent, it seems like adding 2 rules doing it this way which makes it harder to explain why something is like it is.
Is it a mistake or intentional?

geek-soft: Possible approaches:

```
const draft: ResponseDraft = res.locals.user.responseDraft.document
renderView(new Form(draft.document.response), res)
```

or

```
const draft: ResponseDraft = user.responseDraft.document
renderView(new Form(draft.document.response), res)
```

or

```
renderView(new Form(user.responseDraft.document.document.response), res)
```

timja-kainos: I would prefer the third one

timja-kainos: or moving it off the user object

rattur: Yes the third option is better, did a refactor in CCJ check and send route in last sprint

src/main/app/claims/claimMiddleware.ts
```diff
@@ -11,9 +12,10 @@ export class ClaimMiddleware {
     // req.params isn't populated here https://github.com/expressjs/express/issues/2088
     const externalId: string = UUIDUtils.extractFrom(req.path)
 
-    ClaimStoreClient.retrieveByExternalId(externalId, res.locals.user.id)
+    const user: User = res.locals.user
+    ClaimStoreClient.retrieveByExternalId(externalId, user.id)
```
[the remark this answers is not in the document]
geek-soft: I think we should.

geek-soft:
@timja-kainos please have a look again.

PR #219, opened 2017-11-22
Roc 2557 new validation msgs

yarpo:
### JIRA link ###
https://tools.hmcts.net/jira/browse/ROC-2557

### Change description ###
Reworded error messages

```
[ ] Yes
[x] No
```


src/main/features/response/form/models/statement-of-means/dependants.ts
```diff
@@ -6,32 +6,31 @@ import { toNumberOrUndefined } from 'common/utils/numericUtils'
...
   @IsDefined({ message: GlobalValidationErrors.YES_NO_REQUIRED })
   hasAnyChildren: boolean
 
-  @ValidateIf(o => o.hasAnyChildren === true)
-  @IsDefined({ message: ValidationErrors.UNDER_11_REQUIRED })
-  @IsInt({ message: GlobalValidationErrors.NUMBER_REQUIRED })
-  @Min(0, { message: GlobalValidationErrors.NUMBER_REQUIRED })
+  @ValidateIf(o => o.hasAnyChildren === true && !o.under11 && !o.between11and15 && !o.between16and19)
+  @IsInt({ message: ValidationErrors.ENTER_AT_LEAST_ONE })
+  crossFieldError: number
```
geek-soft: That is not very readable. Alternatively we could put some cross field validation in route.

src/main/features/response/form/models/statement-of-means/dependants.ts
```diff
@@ -6,32 +6,31 @@ import { toNumberOrUndefined } from 'common/utils/numericUtils'
...
 
   @IsDefined({ message: GlobalValidationErrors.YES_NO_REQUIRED })
   hasAnyChildren: boolean
 
-  @ValidateIf(o => o.hasAnyChildren === true)
-  @IsDefined({ message: ValidationErrors.UNDER_11_REQUIRED })
-  @IsInt({ message: GlobalValidationErrors.NUMBER_REQUIRED })
-  @Min(0, { message: GlobalValidationErrors.NUMBER_REQUIRED })
+  @ValidateIf(o => o.hasAnyChildren === true && !o.under11 && !o.between11and15 && !o.between16and19)
+  @IsInt({ message: ValidationErrors.ENTER_AT_LEAST_ONE })
```
geek-soft: Do we need this. Usually we make input fields of type `number` and deserialisation method deserialises value to proper type. That solution suggests there expected type is a number but it might not be a number. There is something wrong with the deserialisation.

src/main/features/response/form/models/statement-of-means/dependants.ts
yarpo: I'm my opinion this is better. You see the model class and you know what are rules. I'd keep it in one place.

What we may consider would be different annotation, eg `@AlwaysError` as in fact we do cross field validation in ValidateIf.

Also can be unit tested easily.

src/test/features/response/form/models/statement-of-means/dependants.ts
```diff
@@ -109,41 +109,39 @@ describe('Dependants', () => {
...
 
-            expect(errors.length).to.equal(3)
-            expectValidationError(errors, ValidationErrors.UNDER_11_REQUIRED)
-            expectValidationError(errors, ValidationErrors.BETWEEN_11_AND_15_REQUIRED)
-            expectValidationError(errors, ValidationErrors.BETWEEN_16_AND_19_REQUIRED)
+            expect(errors.length).to.equal(1)
+            expectValidationError(errors, ValidationErrors.ENTER_AT_LEAST_ONE)
           })
 
           it('one of other fields has < 0 given', () => {
```
geek-soft: That only proves `@Min` validation for the first field. What about remaining two?

src/main/features/response/form/models/statement-of-means/dependants.ts
yarpo: It's not about 'type of', it's about 1, 2, 3.. vs 1.1, 3.14 etc :)

geek-soft: I see you trying to prevent numbers with decimal points.

yarpo: It will never be a number as this is a field to reflect cross-field validation errors only. It's not displayed (or even hidden in html code). IMO works good and I have thought about other options (like eg. doing it in routes) and I find it much better - validation is done by model, encapsulated, easier to unit test, tested in the same test file).

src/main/features/response/form/models/statement-of-means/dependants.ts
geek-soft: I appreciate testability but having `crossFieldError: number` for cross browser validation is not readable at all.

src/main/features/response/form/models/statement-of-means/dependants.ts
geek-soft: Solution would be:

1. wrapping numeric values with some internal class
2. implementing custom validator for the internal class ensuring at least one value within instance is populated
3. applying validator to the property of newly created wrapper type

src/main/features/response/form/models/statement-of-means/dependants.ts
```diff
@@ -17,28 +16,13 @@ export class Dependants implements Serializable<Dependants> {
   hasAnyChildren: boolean
 
   @ValidateIf(o => o.hasAnyChildren === true)
-  @IsDefined({ message: ValidationErrors.UNDER_11_REQUIRED })
-  @IsInt({ message: GlobalValidationErrors.NUMBER_REQUIRED })
-  @Min(0, { message: GlobalValidationErrors.NUMBER_REQUIRED })
-  under11: number
+  @ValidateNested({ each: true })
```
geek-soft: I think we do not need to say `{ each: true }` This is just for validating elements of collection while here we have just one instance.

src/main/features/response/form/models/statement-of-means/dependants.ts
```diff
@@ -48,13 +32,11 @@ export class Dependants implements Serializable<Dependants> {
...
       value.hasAnyChildren !== undefined ? toBoolean(value.hasAnyChildren) === true : undefined,
-      toNumberOrUndefined(value.under11),
-      toNumberOrUndefined(value.between11and15),
-      toNumberOrUndefined(value.between16and19)
+      NumberOfChildren.fromObject(value.numberOfChildren)
     )
 
     if (!dependants.hasAnyChildren) {
-      dependants.under11 = dependants.between11and15 = dependants.between16and19 = undefined
+      dependants.numberOfChildren = undefined
```
geek-soft: You can reverse the condition and deserialise `NumberOfChildren` only when `dependants.hasAnyChildren` evaluates to `true`. This way deserialisation method is not called when unneeded.

src/main/features/response/form/models/statement-of-means/maintenance.ts
```diff
@@ -3,17 +3,17 @@ import { IsDefined, IsInt, Min, ValidateIf } from 'class-validator'
...
+  @IsDefined({ message: GlobalValidationErrors.YES_NO_REQUIRED })
   option: boolean
 
   @ValidateIf(o => o.option === true)
-  @IsDefined({ message: ValidationErrors.NUMBER_REQUIRED })
-  @IsInt({ message: ValidationErrors.NUMBER_REQUIRED })
-  @Min(1, { message: ValidationErrors.NUMBER_REQUIRED })
+  @IsDefined({ message: GlobalValidationErrors.NUMBER_REQUIRED })
+  @IsInt({ message: GlobalValidationErrors.INTEGER_REQUIRED })
+  @Min(1, { message: GlobalValidationErrors.POSITIVE_NUMBER_REQUIRED })
```
geek-soft: We usually use `@IsPositive` in such a case which is probably more expressive. Any reason not to use it?

src/main/app/forms/validation/validators/atLeastOneFieldIsPopulated.ts
```diff
@@ -0,0 +1,34 @@
...
+      return true
+    }
+
+    return Object.keys(value).some(key => !!value[key])
+  }
+}
+
+/**
+ * Validates if one of the fields in object is populated. It doesn't matter which one as long as one
+ * as long as at least one of them has any value. It doesn't matter what is the value as long as it's no falsy.
```
geek-soft: This documentation does not read well :(

src/main/features/response/form/models/statement-of-means/dependants.ts
yarpo: ok

src/main/features/response/form/models/statement-of-means/dependants.ts
yarpo: ok

src/main/features/response/form/models/statement-of-means/dependants.ts
yarpo: remove attr

src/main/features/response/form/models/statement-of-means/dependants.ts
yarpo: refactored

src/main/app/forms/validation/validators/atLeastOneFieldIsPopulated.ts
yarpo: reworded. Yeah, this one must have been written by drunken bear ;)

src/test/features/response/form/models/statement-of-means/dependants.ts
```diff
@@ -109,47 +100,75 @@ describe('Dependants', () => {
...
+          })
+
+          it('between11and15 is decimal', () => {
+            const errors = validator.validateSync(new Dependants(true, new NumberOfChildren(0, 1.5, 0)))
+
+            expect(errors.length).to.equal(1)
+            expectValidationError(errors, GlobalValidationErrors.INTEGER_REQUIRED)
+          })
+
+          it('between16and19', () => {
```
geek-soft: Shouldn't it be `between16and19 is decimal`?

PR #222, opened 2017-11-23
ROC-2695 Prevent defendant viewing CCJ pages

timja-kainos:


src/main/app/guards/isClaimantInCaseGuard.ts
```diff
@@ -0,0 +1,21 @@
+import * as express from 'express'
+
+import { GuardFactory } from 'features/response/guards/guardFactory'
+import { ForbiddenError } from '../../errors'
+import { User } from 'idam/user'
+
+export class IsClaimantInCaseGuard {
+  /**
+   * Throws NotFound error user is not defendant in the case
```
geek-soft: Should be `Throws ForbiddenError...`

src/main/app/guards/isClaimantInCaseGuard.ts
```diff
@@ -0,0 +1,21 @@
...
+   *
+   * @returns {express.RequestHandler} - request handler middleware
+   */
+  static check (): express.RequestHandler {
+    return GuardFactory.create((res: express.Response) => {
```
geek-soft: Nice you used the Guard factory so no need to mock it in a11y tests :)

src/main/features/ccj/index.ts
```diff
@@ -22,8 +23,10 @@ function requestHandler (): express.RequestHandler {
 
 export class CCJFeature {
   enableFor (app: express.Express) {
-    app.all(/^\/case\/.+\/ccj\/.*$/, requestHandler())
-    app.all(/^\/case\/.+\/ccj\/.*$/, ClaimMiddleware.retrieveByExternalId)
+    const allCCJ = '/case/*/ccj/*'
```
geek-soft: Great if that works, I do not think it was available to use from the beginning.

PR #224, opened 2017-11-23
ROC-2559: common logic moved to abstract class for multi row forms

yarpo:
### JIRA link ###
https://tools.hmcts.net/jira/browse/ROC-2559

### Change description ###
Abstract class with common logic for multi row forms

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/app/forms/models/itemInMultiRowForm.ts
+++ b/src/main/app/forms/models/itemInMultiRowForm.ts
@@ -1,6 +1,0 @@
-export abstract class ItemInMultiRowForm {
-
-  abstract deserialize (input?: any): ItemInMultiRowForm
-
-  abstract isEmpty (): boolean
-}
--- a/src/main/app/forms/models/itemInMultiRowForm.ts
+++ b/src/main/app/forms/models/itemInMultiRowForm.ts
@@ -1,0 +1,6 @@
+export abstract class ItemInMultiRowForm {
+
+  abstract deserialize (input?: any): ItemInMultiRowForm
+
+  abstract isEmpty (): boolean
+}
--- a/src/main/app/forms/models/multiRowForm.ts
+++ b/src/main/app/forms/models/multiRowForm.ts
@@ -1,23 +1,20 @@
 import { Serializable } from 'models/serializable'
 import { MultiRowFormItem } from 'forms/models/multiRowFormItem'
+import { ValidateNested } from 'class-validator'
+
+export const MAX_NUMBER_OF_ROWS: number = 20
+export const INIT_ROW_COUNT: number = 1
 
 export abstract class MultiRowForm<T extends MultiRowFormItem> implements Serializable<MultiRowForm<T>> {
 
+  @ValidateNested({ each: true })
   rows: T[]
 
-  abstract getMaxNumberOfRows (): number
+  constructor (rows?: T[]) {
+    this.rows = rows || this.initialRows()
+  }
 
   abstract createEmptyRow (): T
-
-  abstract deserializeRows (rows: any): T[]
-
-  deserialize (input?: any): MultiRowForm<T> {
-    if (input) {
-      this.rows = this.deserializeRows(input.rows)
-    }
-
-    return this
-  }
 
   appendRow () {
     if (this.canAddMoreRows()) {
@@ -29,8 +26,42 @@ export abstract class MultiRowForm<T extends MultiRowFormItem> implements Serial
     return this.rows.length < this.getMaxNumberOfRows()
   }
 
+  deserialize (input?: any): MultiRowForm<T> {
+    if (input) {
+      this.rows = this.deserializeRows(input.rows)
+    }
+
+    return this
+  }
+
+  deserializeRows (rows: any): T[] {
+    if (!rows) {
+      return this.initialRows()
+    }
+
+    let employerRows: T[] = rows.map(row => this.createEmptyRow().deserialize(row))
+
+    if (rows.length < this.getInitialNumberOfRows()) {
+      employerRows = employerRows.concat(this.initialRows(this.getInitialNumberOfRows() - rows.length))
+    }
+
+    return employerRows
+  }
+
+  getInitialNumberOfRows (): number {
+    return INIT_ROW_COUNT
+  }
+
+  getMaxNumberOfRows (): number {
+    return MAX_NUMBER_OF_ROWS
+  }
+
   getPopulatedRowsOnly (): T[] {
     return this.rows.filter(item => !item.isEmpty())
+  }
+
+  initialRows (rows?: number): T[] {
+    return new Array(rows || this.getInitialNumberOfRows()).fill(this.createEmptyRow())
   }
 
   removeExcessRows () {
--- a/src/main/app/forms/models/multiRowForm.ts
+++ b/src/main/app/forms/models/multiRowForm.ts
@@ -39,13 +39,13 @@ export abstract class MultiRowForm<T extends MultiRowFormItem> implements Serial
       return this.initialRows()
     }
 
-    let employerRows: T[] = rows.map(row => this.createEmptyRow().deserialize(row))
+    let deserialisedRows: T[] = rows.map(row => this.createEmptyRow().deserialize(row))
 
     if (rows.length < this.getInitialNumberOfRows()) {
-      employerRows = employerRows.concat(this.initialRows(this.getInitialNumberOfRows() - rows.length))
+      deserialisedRows = deserialisedRows.concat(this.initialRows(this.getInitialNumberOfRows() - rows.length))
     }
 
-    return employerRows
+    return deserialisedRows
   }
 
   getInitialNumberOfRows (): number {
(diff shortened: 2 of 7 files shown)


src/main/app/forms/models/itemInMultiRowForm.ts
```diff
@@ -0,0 +1,6 @@
+export abstract class ItemInMultiRowForm {
```
rlewan: `MultiRowFormItem`?

rlewan: Or `FormRow`?

radoslawn-kainos: Why an abstract class instead of an interface?

src/main/app/forms/models/itemInMultiRowForm.ts
```diff
@@ -0,0 +1,6 @@
...
+
+  abstract deserialize (input?: any): ItemInMultiRowForm
```
radoslawn-kainos: There's no reason to specify this method here. Any class extending MultiRowForm has to know how to create its items anyway.

src/main/app/forms/models/multiRowForm.ts
```diff
@@ -0,0 +1,43 @@
+import { Serializable } from 'models/serializable'
+import { ItemInMultiRowForm } from 'forms/models/itemInMultiRowForm'
+
+export abstract class MultiRowForm<T extends ItemInMultiRowForm> implements Serializable<MultiRowForm<T>> {
```
radoslawn-kainos: Serializable doesn't serve any purpose. It's not being explicitly consumed anywhere. To be honest, I think it's best if we stop using it. It doesn't even do what the name suggests.

radoslawn-kainos: If this is meant to be reusable, I think it would make sense to have some unit tests for it.

src/main/app/forms/models/multiRowForm.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  rows: T[]
```
radoslawn-kainos: If we're making an abstraction, why require the rows to be stored in any particular way? Maybe a getter method returning a ReadonlyArray would be more appropriate?

src/main/app/forms/models/multiRowForm.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  abstract getMaxNumberOfRows (): number
+
+  abstract createEmptyRow (): T
+
+  abstract deserializeRows (rows: any): T[]
+
+  deserialize (input?: any): MultiRowForm<T> {
```
radoslawn-kainos: Why does it need to be an instance method instead of a static one?

src/main/app/forms/models/itemInMultiRowForm.ts
yarpo: I'd rather call it `MultiRowFormItem`. More descriptive imo.

src/main/app/forms/models/multiRowForm.ts
yarpo: This is just the design we already have in 100% of our models. 

I my humble opinion this could (should) be static. BUT as we are where we are I'd not change it now. 0 (zero) business value a lot of work.

yarpo: @geek-soft

yarpo:
@radoslawn-kainos not at the moment. But there is a story to create a helper for all check and send pages so that we can reuse it on all c'ns pages. Then we would use could use the common interface

src/main/features/response/form/models/statement-of-means/employers.ts
```diff
@@ -28,33 +28,11 @@ export class Employers implements Serializable<Employers> {
...
+  getMaxNumberOfRows (): number {
+    return MAX_NUMBER_OF_JOBS
   }
 
-  canAddMoreRows () {
-    return this.rows.length < MAX_NUMBER_OF_JOBS
-  }
-
-  private deserializeRows (rows: any): EmployerRow[] {
+  deserializeRows (rows: any): EmployerRow[] {
```
geek-soft: I would remove this to base class as well. Deserialisation logic will be almost the same for all cases.

src/main/features/response/form/models/statement-of-means/employers.ts
```diff
@@ -68,4 +46,7 @@ export class Employers implements Serializable<Employers> {
     return employerRows
   }
 
+  createEmptyRow (): EmployerRow {
```
geek-soft: That belongs to `EmployerRow`. We can replace by requiring default contractor.

src/main/features/response/form/models/statement-of-means/employers.ts
```diff
@@ -28,33 +28,11 @@ export class Employers implements Serializable<Employers> {
...
-    }
-  }
-
-  removeExcessRows () {
-    this.rows = this.rows.filter(item => !!item.jobTitle && !!item.employerName)
-
-    if (this.rows.length === 0) {
-      this.appendRow()
-    }
+  getMaxNumberOfRows (): number {
```
geek-soft: That could be just property getter that can be overridden instead of forcing method syntax.

src/main/app/forms/models/multiRowFormItem.ts
```diff
@@ -0,0 +1,12 @@
+export abstract class MultiRowFormItem {
+
+  abstract deserialize (input?: any): MultiRowFormItem
+
+  isEmpty (): boolean {
+    const that = this
```
radoslawn-kainos: You don't have to do this when using arrow function syntax

src/main/app/forms/models/multiRowFormItem.ts
```diff
@@ -0,0 +1,12 @@
+export abstract class MultiRowFormItem {
+
+  abstract deserialize (input?: any): MultiRowFormItem
+
+  isEmpty (): boolean {
```
radoslawn-kainos: Have you tried something like:
```ts
return Object.values(this).every(value => !value)
```
or alternatively:
```ts
return !Object.values(this).some(value => value)
```

rlewan: What is this function supposed to do? Check that inputs are blank?

src/main/app/forms/models/multiRowForm.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    return this
+  }
+
+  deserializeRows (rows: any): T[] {
+    if (!rows) {
+      return this.initialRows()
+    }
+
+    let employerRows: T[] = rows.map(row => this.createEmptyRow().deserialize(row))
```
rlewan: reference to `employer` in an abstract funtion.

src/main/app/forms/models/multiRowForm.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  initialRows (rows?: number): T[] {
+    return new Array(rows || this.getInitialNumberOfRows()).fill(this.createEmptyRow())
+  }
+
+  removeExcessRows () {
+    this.rows = this.getPopulatedRowsOnly()
+
+    if (this.rows.length === 0) {
+      this.appendRow()
```
rlewan: Why append only one row in this case?

rlewan:
@radoslawn-kainos 
> Overall, I'm not convinced if inheritance is a good way to go here. Is the base class going to be directly consumed anywhere in the code? Maybe extracting commonly shared behaviour to some helper functions is all we need?

> I don't like the approach with abstract classes and inheritance, as I feel a generic collection and composition would be clearer and more maintainable, but I guess I'm outnumbered, so I give up.

If we were to extract this logic to helper functions, I think we would then have have a number functions which all need a `rows` input parameter and which then mutate this input parameter (which is not great if we're talking about clean code here). Or we would need to create clones each time and return those to avoid mutation.

I see your point that using inheritance for code re-use only and not for polymorphism can be a code smell indicator but I think it's fine as Patryk implemented it, it feels alright that these functions are a part of multi row and row classes. Also, it's hard for me to imagine that `MultiRowForm` and `MultiRowFormItem` would need to enter additional inheritance hierarchies and this code would get in the way.

src/main/app/forms/models/multiRowFormItem.ts
yarpo: All fields are empty (value is falsy: undefined / '')

radoslawn-kainos:
@rlewan 
Yeah, I kind of get it, but if you want to know my opinion, [this sums it up nicely](https://xkcd.com/292/).
To me, for something to be a good design decision, it's not enough that it doesn't create any problems you can think of at the moment. Good practices and principles exist so that we can deny whole classes of problems the chance to appear, instead of having to think of every possible consequence when making every choice.

In this particular case: Let's say we adopt this pattern, and start extracting other common behaviours the same way. What happens if we encounter a page that has more than one of them? Multiple inheritance? Even if it was possible, I doubt it would be a good idea. Am I missing something here? Of course, we could just go this way and change it to something else when needed.

As for the solution I had in mind, yeah, I mentioned some helpers first, but then thought about making it a generic container class (to be composed into the form model, not extended by it), at the potential cost of one more level of nesting of the list of rows within the whole form model. I think an example would illustrate it best, so I could try to come up with something on Monday. On the other hand, since this is almost done and most of you seem to agree that it's fine, so is it worth the time?

PR #228, opened 2017-11-28
ROC-2243: Fix accessibility issues

radoslawn-kainos:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-2243

### Work checklist

- [ ] Yes
- [x] No

### DIFF
--- a/src/main/features/claim/views/reason.njk
+++ b/src/main/features/claim/views/reason.njk
@@ -7,20 +7,36 @@
 {% block content %}
   <div class="grid-row">
     <div class="column-two-thirds">
-
-      <p>{{ t('Briefly explain why you are owed the money.') }}</p>
-      <div class="panel panel-border-wide">
-        <p>{{ t('Your claim will be sent to the defendant %1$s - a judge will not consider your explanation unless the defendant %1$s rejects your claim.',
-        {
-          postProcess: 'sprintf',
-          sprintf: [defendantName]
-        })
-        }}</p>
-      </div>
       <form novalidate method="post">
         {{ csrfProtection(csrf) }}
 
-        {{ textArea(label = '', name = 'reason', form = form, rows = 10) }}
+        {% set name = 'reason' %}
+        {% set error = form.errorFor(name) %}
+        <div class="form-group {% if error %} form-group-error {% endif %}">
+          <label for="{{ name }}" id="{{ name }}[label]">
+            <span class="form-label">
+              {{ t('Briefly explain why you are owed the money.') }}
+            </span>
+
+            <div class="panel panel-border-wide">
+              <p>{{
+                t('Your claim will be sent to the defendant {{defendantName}} - a judge will not consider your explanation unless the defendant {{defendantName}} rejects your claim.',
+                  {
+                    postProcess: 'sprintf',
+                    sprintf: [defendantName]
+                  }
+                )
+              }}</p>
+            </div>
+
+            {% if error %}
+              <span class="error-message">{{ t(error) }}</span>
+            {% endif %}
+          </label>
+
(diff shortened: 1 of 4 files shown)


src/main/features/claim/views/reason.njk
```diff
@@ -7,20 +7,36 @@
...
+        {% set error = form.errorFor(name) %}
+        <div class="form-group {% if error %} form-group-error {% endif %}">
+          <label for="{{ name }}" id="{{ name }}[label]">
+            <span class="form-label">
+              {{ t('Briefly explain why you are owed the money.') }}
+            </span>
+
+            <div class="panel panel-border-wide">
+              <p>{{
+                t('Your claim will be sent to the defendant {{defendantName}} - a judge will not consider your explanation unless the defendant {{defendantName}} rejects your claim.',
```
timja-kainos: can you add some spacing around `{{defendantName}}` please?

src/main/features/claim/views/reason.njk
```diff
@@ -7,20 +7,36 @@
...
+                  {
+                    postProcess: 'sprintf',
+                    sprintf: [defendantName]
```
timja-kainos: does this work? usually we do %s and not the variable name directly?

src/main/features/claim/views/reason.njk
```diff
@@ -7,20 +7,36 @@
...
+                )
+              }}</p>
+            </div>
+
+            {% if error %}
+              <span class="error-message">{{ t(error) }}</span>
+            {% endif %}
+          </label>
+
+          <textarea id="{{ name }}" name="{{ name }}" rows="5" class="form-control form-control-3-4 {% if error %}form-control-error{% endif %}">{{ form.valueFor(name) | default('') }}</textarea>
```
timja-kainos: can we use the text area macro? / fix the macro if there's something wrong with it?

src/main/features/claim/views/reason.njk
radoslawn-kainos: Good catch - I tried to replace it with interpolation first, but when I couldn't get it to work the first time, I wanted to reverted to sprintf, but for some reason I missed that.

src/main/features/claim/views/reason.njk
radoslawn-kainos: The problem I see with these macros, especially shared ones like this, is that it's hard to predict the impact when changing it. What other projects are using them? Combined with that their responsibility seems to be poorly defined (are they supposed to render labels too, or just the field, what exactly do we want to be able to configure with arguments, etc.), I preferred not to touch them.

Also, if you know any way to pass html to macros in a way that wouldn't make you lose self-respect as a developer, I'm all ears.

timja-kainos: just had a look and the textArea macro in doing too much, it should just be doing something like this:
https://github.com/hmcts/cmc-common-frontend/blob/master/macros/form.njk#L19

timja-kainos: the only other project that could be using the macros is:
https://github.com/hmcts/cmc-legal-rep-frontend
I've just checked and its being used in 3 places:
```
$ rg textArea
src/main/features/certificateOfService/views/what-documents.njk
6:{% from "form.njk" import csrfProtection, textArea, checkboxOption, saveAndContinueButton, errorSummary %}
68:          {{ textArea(t('List them here:'), 'otherDocuments', form) }}

src/main/features/claim/views/summarise-the-claim.njk
2:{% from "form.njk" import csrfProtection, saveAndContinueButton, errorSummary, textArea %}
21:        {{ textArea(label = '', name = 'text', form = form, inputClass='full-width-textarea', rows = 10) }}

src/main/features/claim/views/what-type-of-claimant.njk
6:{% from "form.njk" import csrfProtection, saveAndContinueButton, textArea, textInput, errorSummary  %}
39:          {{ textArea(t('Full name'), 'claimant_name', form) }}
```

The benefit of trying to fix the macro means that all pages using it get the accessibility fixes, rather than doing them one at a time. Would be great if you can see a way to solve it.

PR #244, opened 2017-12-01
Toggle cluster mode

rlewan:
### Change description

Cluster mode has to be toggled as IIS does not support that.

- [ ] Yes
- [x] No


package.json
```diff
@@ -9,7 +9,7 @@
     "compile": "tsc --project tsconfig.prod.json",
     "copy-assets": "gulp sass copy-files",
     "setup": "yarn compile && yarn copy-assets",
-    "start": "node -r tsconfig-paths/register src/main/server.js",
+    "start": "node src/main/server.js",
```
timja-kainos: this is a separate change but also needed for CNP +1

src/main/server.ts
```diff
@@ -33,26 +36,20 @@ function getSSLOptions (): any {
   }
 }
 
-function forkListenerProcesses (numberOfCores: number) {
-  for (let i = 0; i < numberOfCores; i++) {
-    cluster.fork()
+function applicationPort (): any {
+  let port: any = process.env.PORT
```
timja-kainos: `process.env.PORT || defaultPort` ? 
Is the log message about using the default port still useful?

rlewan: I think it is.

rlewan: Proved itself at least once today 📦

src/main/server.ts
```diff
@@ -33,26 +36,20 @@ function getSSLOptions (): any {
   }
 }
 
-function forkListenerProcesses (numberOfCores: number) {
-  for (let i = 0; i < numberOfCores; i++) {
-    cluster.fork()
+function applicationPort (): any {
```
timja-kainos: probably should have a comment about why this is `any` and if you have time on Monday would be awesome to look at sending a PR for the express types

src/main/server.ts
timja-kainos: but every time a developer starts up their local?

Jenkinsfile
```diff
@@ -1,168 +0,0 @@
-#!groovy
```
geek-soft: I am sure you don't want to delete this just yet.

src/main/server.ts
```diff
@@ -33,26 +36,20 @@ function getSSLOptions (): any {
...
+    port = defaultPort
   }
-  cluster.on('exit', (worker, code, signal) => {
-    logger.info(`Worker ${worker.process.pid} exited with ${code ? code : signal}`)
-  })
+  return port
 }
 
-const testingSupportActivated = toBoolean(config.get<boolean>('featureToggles.testingSupport'))
+const entryPoint = () => listen(app, applicationPort())
```
geek-soft: What about `startApplicationFn`, then we have `cluster.run(startApplicationFn)` and `startApplicationFn()`.

The other idea I had was to extract `run()` method to interface and provide two implementations, one which just does `startApplicationFn()` and other one which does the cluster run.  This way in the code we would have if cluster is disabled:

```
ApplicationRunner.run(startApplicationFn)
```

and when cluster enabled

```
ClusteredApplicationRunner.run(startApplicationFn)
```

src/main/server.ts
```diff
@@ -33,26 +36,20 @@ function getSSLOptions (): any {
   }
 }
 
-function forkListenerProcesses (numberOfCores: number) {
-  for (let i = 0; i < numberOfCores; i++) {
-    cluster.fork()
+function applicationPort (): any {
+  let port: any = process.env.PORT
+  if (port === undefined) {
+    logger.info(`Port value was not set using PORT env variable, using the default of ${defaultPort}`)
```
geek-soft: Does it really matter for us how that was configured?

src/main/server.ts
geek-soft: If we cannot restrict this to number alone lets do `number | string`.

src/main/server.ts
```diff
@@ -1,17 +1,20 @@
...
 import * as path from 'path'
 import * as https from 'https'
-import * as cluster from 'cluster'
+
+import { app } from './app'
+import { ApplicationCluster } from './applicationCluster'
 
 const logger = require('@hmcts/nodejs-logging').getLogger('server')
 
+const defaultPort = 3000
```
geek-soft: Default port should be close to the place it is needed, Please move it to the method that returns application port.

Other thing, since this is more complex shall we extract logic to separate class and make it something like `PortUtil` or `PortFinder`?

Jenkinsfile
timja-kainos: hah nope, `git checkout master Jenkinsfile` will fix it

src/main/server.ts
timja-kainos: looks like this works now, I'm pretty sure before it had to be a number or any.

rlewan: Why is it a problem?

src/main/server.ts
rlewan: Yes, will add a comment and see about the PR.

src/main/server.ts
rlewan: It matters for Azure deployments where the port _has_ to be set through the variable. Not logging this out may make troubleshooting harder.

src/main/applicationCluster.ts
```diff
@@ -0,0 +1,27 @@
...
+    if (cluster.isMaster) {
+      logger.info(`Master process running on ${process.pid}`)
+      const numberOfCores = cpus().length
+      this.forkListenerProcesses(numberOfCores)
+    } else {
+      applicationEntryPoint()
+    }
+  }
+
+  private static forkListenerProcesses (numberOfCores: number) {
```
timja-kainos: void ?

src/main/applicationRunner.ts
```diff
@@ -0,0 +1,44 @@
...
+    }
+  }
+
+  /**
+   * Return type is number or string because Azure PaaS communicates with the application
+   * through a named pipe and not a TCP port.
+   */
+  private static applicationPort (): number | string {
+    const defaultPort = 3000
+    let port: number | string = process.env.PORT
```
timja-kainos: this won't return a number its always a string from an env variable

rlewan: Fair point, will update.

PR #253, opened 2017-12-06
ROC-2772: reuse multi row form class, updated check and send page

yarpo:
### Change description
Refactoring multi row forms so that they all use the same base class and implement only part of logic related to the particular model.

### Work checklist

- [ ] Yes
- [x] No


src/main/app/forms/models/multiRowFormItem.ts
```diff
@@ -5,4 +5,8 @@ export abstract class MultiRowFormItem {
   isEmpty (): boolean {
     return Object.values(this).every(value => !value)
   }
+
+  isAtLeastOneFieldPopulated (): boolean {
```
radoslawn-kainos: minor naming suggestion: `isAnyFieldPopulated`

EDIT: Now that I think about it, shouldn't it be an exact opposite of `isEmpty`? And given that, why not just use `!row.isEmpty()` everywhere instead of this?

src/test/features/response/form/models/timeline.ts
```diff
@@ -11,7 +11,7 @@ describe('Timeline', () => {
 
       const actual: TimelineRow[] = (new Timeline()).rows
 
-      expect(actual.length).to.equal(4)
+      expect(actual.length).to.equal(INIT_ROW_COUNT)
```
radoslawn-kainos: Fix: what if INIT_ROW_COUNT is changed to be something different than 4? Test description is explicit that it should be 4, so that's what test should be checking for. Alternatively, the description needs to be changed.

src/test/features/response/form/models/timeline.ts
```diff
@@ -118,11 +118,11 @@ describe('Timeline', () => {
 
       expect(actual.rows.length).to.be.eq(INIT_ROW_COUNT)
 
-      for (let i = 0; i < MAX_NUMBER_OF_EVENTS + 1; i++) {
+      for (let i = 0; i < MAX_NUMBER_OF_ROWS + 1; i++) {
         actual.appendRow()
       }
 
-      expect(actual.rows.length).to.be.eq(MAX_NUMBER_OF_EVENTS)
+      expect(actual.rows.length).to.be.eq(MAX_NUMBER_OF_ROWS)
```
radoslawn-kainos: Same here - test checks for something different (MAX_NUMBER_OF_ROWS) than it says (20). In this case, it seems that the inconsistency was already there, so it's up to you to decide it you want to fix it.

src/main/features/response/views/statement-of-means/check-and-send.njk
```diff
@@ -109,12 +109,12 @@
 {% endfor %}
 
 {{ row('Court orders', '', statementOfMeansPaths.courtOrdersPage.evaluateUri({ externalId: claim.externalId }), false) }}
-{{ row('Has a court ordered you to pay anyone else?', 'Yes' if draft.statementOfMeans.courtOrders.hasAnyCourtOrders else 'No', '', not draft.statementOfMeans.courtOrders.hasAnyCourtOrders) }}
+{{ row('Are you paying any other court orders?', 'Yes' if draft.statementOfMeans.courtOrders.hasAnyCourtOrders else 'No', '', not draft.statementOfMeans.courtOrders.hasAnyCourtOrders) }}
```
timja-kainos: these content changes look separate, could they go in a separate PR?

src/test/features/response/form/models/timeline.ts
yarpo: done

src/test/features/response/form/models/timeline.ts
yarpo: done

PR #255, opened 2017-12-08
ROC-2826: Run unit tests only coverage on pull requests

geek-soft:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-2826

### Change description

This change should only report coverage from unit tests on pull requests. That should let us focus on unit tests more as running coverage with route tests always reported very hight coverage due to the nature of route tests.

### Work checklist

- [ ] Yes
- [x] No


.travis.yml
```diff
@@ -3,12 +3,12 @@ cache: yarn
 script:
   - yarn --version
   - yarn check --integrity
-  - yarn setup
+  - yarn copy-assets
```
geek-soft: Presence of compiled JS files in the same location as TS files broke coverage report. We do not need to compile TS into JS on travis so this is quick fix. We might need to look into that issue carefully very soon as this is another issue caused by having TS and JS in the same place.

timja-kainos: doesn't the same problem happen on jenkins with sonar? and it will be happening on CNP as well I assume @rlewan ?

package.json
```diff
@@ -16,9 +16,10 @@
     "test:routes": "NODE_ENV=mocha LOG_LEVEL=OFF mocha --opts mocha.opts $(find src/test \\( -name '*.ts' \\) -path '*/routes/*') --timeout 4000 --reporter-options reportFilename=routes,inlineAssets=true,reportTitle=citizen-frontend-routes",
     "test:a11y": "NODE_ENV=mocha LOG_LEVEL=ERROR mocha --exit --opts mocha.opts src/test/a11y/a11y.ts --timeout 60000 --reporter-options reportFilename=a11y,inlineAssets=true,reportTitle=citizen-frontend-a11y",
     "test:nsp": "nsp check",
-    "test:coverage": "NODE_ENV=mocha LOG_LEVEL=OFF nyc mocha --opts mocha.opts $(find src/test \\( -name '*.ts' -or -name '*.js' \\) ! -path '*a11y*')",
+    "test:coverage": "NODE_ENV=mocha LOG_LEVEL=OFF nyc mocha --opts mocha.opts $(find src/test \\( -name '*.ts' \\) ! -path '*a11y*')",
+    "test:coverage-unit": "NODE_ENV=mocha LOG_LEVEL=OFF nyc mocha --opts mocha.opts $(find src/test \\( -name '*.ts' \\) ! -path '*a11y*' ! -path '*/routes/*')",
     "test:codacy-upload": "cat ./coverage-report/lcov.info | codacy-coverage",
-    "test:codecov-upload": "nyc report --reporter=json && codecov -f coverage-report/*.json",
+    "test:codecov-upload": "codecov -f ./coverage-report/lcov.info",
```
timja-kainos: I could have sworn codecov required json, nice not to have to convert it 👍

.travis.yml
geek-soft: I removed TS compilation from Jenkinsfile as well. You were right @timja-kainos , the same issue was with sonar reporting lower coverage.

rlewan: Yes, sonar was scanning `*.js` files.

timja-kainos: @rlewan Does this need to be changed in CNP branch as well so that it is 
```
afterCheckout
yarn copy-assets

afterSonar (something like that)
yarn compile
```

geek-soft: I would postpone changes to CNP till we decide whether we need to compile into JS. When https://tools.hmcts.net/jira/browse/ROC-2908 is played I plan to either remove TS>JS compilation and stick back to ts-node (if it is fast enough) or use separate TS>JS compilation target directory.

timja-kainos: 👍

PR #257, opened 2017-12-11
ROC-2005 Postcode lookup

timja-kainos:
https://tools.hmcts.net/jira/browse/ROC-2005


src/main/public/js/postcode-lookup.js
```diff
@@ -0,0 +1,134 @@
...
+          showAddressEntry(postcodeLookupWidget)
+        })
+
+      if (isAnyAddressFieldPopulated(postcodeLookupWidget)) {
+        showAddressEntry(postcodeLookupWidget)
+      }
+    })
+})
+
+function isNorthernIrelandPostcode (postcode) {
```
yarpo: do we want these functions to be global?

src/main/public/js/postcode-lookup.js
```diff
@@ -0,0 +1,134 @@
...
+    northernIrelandErrorMessage.classList.add('hidden')
+    genericErrorMessage.classList.remove('hidden')
+  }
+
+  postcodeLookupWidget.querySelector('.postcode-search-container').classList.add('form-group-error')
+}
+
+function lookupPostcode (postcode, postcodeLookupWidget) {
+  var xhr = new XMLHttpRequest()
+  xhr.open('GET', '/postcode-lookup?postcode=' + encodeURIComponent(postcode))
```
yarpo: maybe we can add some object to handle Ajax requests? probably there will be more and more scripts like this one?

yarpo: eg:
https://github.com/yarpo/yarpobject/blob/master/ajax/yAjax/yajax.js

timja-kainos: no, YAGNI

package.json
```diff
@@ -22,12 +22,13 @@
...
   },
   "dependencies": {
-    "@hmcts/cmc-common-frontend": "^1.29.0",
+    "@hmcts/cmc-common-frontend": "^1.31.0",
     "@hmcts/cmc-draft-store-middleware": "^1.0.0",
     "@hmcts/draft-store-client": "^1.1.1",
     "@hmcts/info-provider": "^1.0.0",
     "@hmcts/nodejs-healthcheck": "^1.4.5",
     "@hmcts/nodejs-logging": "^1.4.2",
+    "@hmcts/postcodeinfo-client": "^0.0.1",
```
robertparkinson: shouldn't this be 1.0.0 ?

src/main/public/js/postcode-lookup.js
```diff
@@ -0,0 +1,158 @@
...
+      }
+    })
+
+  document.querySelector('#saveAndContinue')
+    .addEventListener('click', function (event) {
+      document.querySelectorAll('.postcode-lookup')
+        .forEach(function(postcodeLookupWidget) {
+          var addressInputsVisible = !postcodeLookupWidget.querySelector('.address')
+            .classList.contains('js-hidden')
+          console.log('address', addressInputsVisible)
```
robertparkinson: should this console.log be in here?

src/main/public/js/postcode-lookup.js
```diff
@@ -0,0 +1,158 @@
...
+          postcodeLookupWidget.querySelector('.postcode-address-visible').value = addressInputsVisible
+
+
+          var addressPickerVisible = !postcodeLookupWidget.querySelector('.postcode-address-picker')
+            .classList.contains('hidden')
+          console.log('picker', addressPickerVisible)
```
robertparkinson: and here?

src/main/public/js/postcode-lookup.js
```diff
@@ -0,0 +1,158 @@
...
+      var ni = isNorthernIrelandPostcode(postcode)
+      showAddressError(ni, postcodeLookupWidget)
+      showAddressEntry(postcodeLookupWidget)
+      return
+    }
+
+    var postcodeSelectDropdown = postcodeLookupWidget.querySelector('select')
+
+    var nonSelectableOption = document.createElement("option")
+    nonSelectableOption.text = postcodeResponse.addresses.length + ' addresses found'
```
robertparkinson: how are we going to handle translating this?

rlewan:
Do we want to toggle this feature?

package.json
rlewan: +1, should be `1.0.0` or `0.1.0`.

src/main/public/js/postcode-lookup.js
```diff
@@ -0,0 +1,158 @@
...
+  addressSelector.querySelector('.address-town-or-city').value = ''
+  addressSelector.querySelector('.postcode').value = ''
+}
+
+function isAnyAddressFieldPopulated(postcodeLookupWidget) {
+  var addressSelector = postcodeLookupWidget.querySelector('.address')
+
+  return addressSelector.querySelector('.address-line1').value !== '' ||
+    addressSelector.querySelector('.address-line2').value !== '' ||
+    addressSelector.querySelector('.address-town-or-city').value !== '' ||
```
rlewan: I can see plenty of `addressSelector.querySelector` calls, some of them to the same DOM element. What do you think about extracting at least some of them and then have e.g. `addressLine2().value`?

src/main/public/js/postcode-lookup.js
```diff
@@ -0,0 +1,158 @@
...
+  postcodeLookupWidget.querySelector('.postcode-search-container').classList.add('form-group-error')
+}
+
+function lookupPostcode (postcode, postcodeLookupWidget) {
+  var xhr = new XMLHttpRequest()
+  xhr.open('GET', '/postcode-lookup?postcode=' + encodeURIComponent(postcode))
+  xhr.onload = function () {
+    if (xhr.status !== 200) {
+      showAddressError(false, postcodeLookupWidget)
+      showAddressEntry(postcodeLookupWidget)
```
rlewan: What do you think about adding `show(element)` and `hide(element)` functions? Then, combined with my other comment on this file` we could have e.g. `show(postcodeLookupWidget())`.

src/main/public/js/postcode-lookup.js
rlewan: Well spotted @robertparkinson. I suppose this has to be hidden content on the page, translated server-side?

rlewan: Either a hidden span you hide and show, or just some hidden element which acts a placeholder for text which you extract with client-side JS.

src/main/routes/postcode-lookup.ts
```diff
@@ -0,0 +1,26 @@
...
+import * as Logging from '@hmcts/nodejs-logging'
+
+const postcodeClient = new PostcodeInfoClient(config.get<string>('postcodeLookup.apiKey'), request)
+const logger = Logging.getLogger('postcode-lookup')
+
+/* tslint:disable:no-default-export */
+export default express.Router()
+  .get(AppPaths.postcodeLookupProxy.uri, (req, res) => {
+    postcodeClient.lookupPostcode(req.query.postcode)
+      .then((postcodeInfoResponse: PostcodeInfoResponse) => res.json(postcodeInfoResponse))
```
rlewan: What do you think about using `async`/`await` for consistency?

src/test/routes/mocks/mockPostcodeInfoClient.ts
```diff
@@ -0,0 +1,16 @@
+import * as mock from 'mock-require'
+
+mock('@hmcts/postcodeinfo-client', './mockPostcodeInfoClient')
+
+export class PostcodeInfoClient {
+  constructor (public readonly apikey: string, public readonly request?, apiLocation?) {
```
rlewan: I think these properties can be private?

src/main/public/js/postcode-lookup.js
timja-kainos: showing and hiding is a bit specific depending on how its hidden because we have to support a non js fall back, there's `js-hidden` `hidden` and `js-visible` not sure if it could be made generic

src/main/app/forms/models/address.ts
```diff
@@ -25,21 +26,46 @@ export class ValidationConstants {
 
 export class Address implements CompletableTask {
 
+  @ValidateIf(o => o.addressVisible)
   @IsDefined({ message: ValidationErrors.FIRST_LINE_REQUIRED, groups: ['claimant', 'defendant', 'response'] })
   @IsNotBlank({ message: ValidationErrors.FIRST_LINE_REQUIRED, groups: ['claimant', 'defendant', 'response'] })
-  @MaxLength(ValidationConstants.ADDRESS_MAX_LENGTH, { message: ValidationErrors.FIRST_LINE_TOO_LONG, groups: ['claimant', 'defendant', 'response'] })
+  @MaxLength(ValidationConstants.ADDRESS_MAX_LENGTH, {
+    message: ValidationErrors.FIRST_LINE_TOO_LONG,
```
yarpo: why not `(global)ValidationErrors.TEXT_TOO_LONG`

src/main/routes/postcode-lookup.ts
timja-kainos: I think each have their part and should be used as needed, I don't think changing it gives any value.

src/main/app/forms/models/address.ts
```diff
@@ -55,13 +81,12 @@ export class Address implements CompletableTask {
     if (input == null) {
       return input
     }
-    let deserialized: Address = new Address(
-        input.line1,
-        input.line2,
-        input.city,
-        input.postcode
+    return new Address(
```
yarpo: how other fields are populated?

src/main/routes/postcode-lookup.ts
robertparkinson: I actually prefer the .then syntax as was thinking of suggesting we move to it although that would be quite a big change

src/main/public/js/postcode-lookup.js
yarpo: If they are always in DOM I would even assign reference to DOM node and use var:
```
var element = addressSelector.querySelector('.address-line2')
element.value = ...
```

src/test/routes/mocks/mockPostcodeInfoClient.ts
timja-kainos: typescript wouldn't compile if I had them private and didn't use them :(

src/main/app/forms/models/address.ts
timja-kainos: didn't change that

src/main/app/forms/models/address.ts
timja-kainos: hmm?

src/main/public/js/postcode-lookup.js
yarpo: Maybe this should be in some config?

```
/postcode-lookup?postcode=
```

src/main/public/stylesheets/components/_form.scss
```diff
@@ -65,3 +65,7 @@ select {
     width: 50%;
   }
 }
+
+.second-text-input {
+  margin-top: 15px;
```
yarpo: I can see a lot of places where we have margin-*: 15px. Why don't we add some sccs var for this value?

NOT FOR THIS PR

src/main/public/js/postcode-lookup.js
rlewan: Don't bother if it would overcomplicate the code.

src/test/routes/postcode-lookup.ts
```diff
@@ -0,0 +1,26 @@
...
+  describe('on GET', () => {
+    it('should return result from postcode client', () =>
+      request(app)
+        .get(`${AppPaths.postcodeLookupProxy.uri}?postcode=111`)
+        .set('Cookie', `${cookieName}=ABC`)
+          .expect(200, { valid: false })
+    )
+    it('should return error from postcode client', () =>
+      request(app)
+        .get(`${AppPaths.postcodeLookupProxy.uri}?postcode=fail`)
```
yarpo:
```
const apiUrl = AppPaths.postcodeLookupProxy.uri
```

and use this const twice in the code?

timja-kainos:
I would say no to toggling why would we need to?

src/main/public/js/postcode-lookup.js
timja-kainos: why? we control that in the frontend. if someone needs to change it they can.
I don't see any value in extracting that out and a lot of extra unneeded work to load the config from the backend and extra ajax calls etc

rlewan:
RE toggling -- I thought it's a new feature so just to be on the safe side in case testing finds some issues. But I suppose release date is far away so we have enough time to act.

src/test/routes/postcode-lookup.ts
timja-kainos: done

src/main/routes/postcode-lookup.ts
rlewan: @robertparkinson we were using the promise API syntax initially and then decided to move to async/await, that's the reason for my comment initial comment on this line :)

src/main/public/js/postcode-lookup.js
timja-kainos: @yarpo I don't think any of them are in scope for that to work and give value,
I'll have a look at what rafal suggested after I have the tests fixed

PR #258, opened 2017-12-11
Fix ts-node restarting

timja-kainos:


timja-kainos:
may not actually be working

geek-soft:
Closing this as it did not solve the problem for everyone and is not needed temporarily due to the fact ts-node upgrade got reverted.

PR #268, opened 2017-12-13
Use ts-node fast mode to speed up compilation time

geek-soft:
### JIRA link

???

### Change description

This PR enables ts-node fast mode. That speeds up performance so that compilation with empty cache takes around 7 seconds on my backup machine (slower then yours :P).

I tried ts-node 4.0.1 however it looks like compilation time when that version is used is slightly but noticeable higher. I have impression that ts-node 4.0.1 does not use cache properly which makes each compilation start with empty cache. I need to investigate it more deeply before I raise that issue with ts-node team.

### Work checklist

- [ ] Yes
- [x] No


tsconfig.json
```diff
@@ -10,7 +10,7 @@
       "es2017.object",
       "dom"
     ],
-    "target": "es5",
+    "target": "es6",
```
geek-soft: Not strictly necessary however ES6 code is smaller then ES5 (2.8MB vs 3.1MB) so your hard drives won't work that hard and you probably save 1ms here :)

geek-soft:
Either as a part of this PR or in separate PR I will remove remove TS>JS compilation. It will make startup time longer by around 4 seconds on my machine but we will get proper stack traces back.

package.json
```diff
@@ -10,14 +10,14 @@
     "copy-assets": "gulp sass copy-files",
     "setup": "yarn compile && yarn copy-assets",
     "start": "node src/main/server.js",
-    "start-dev": "./node_modules/.bin/ts-node src/main/server.ts",
+    "start-dev": "TS_NODE_FAST=true ./node_modules/.bin/ts-node src/main/server.ts",
```
timja-kainos: not planning on upgrading back to 4.0?

geek-soft: Not now due to the fact that event if we did not have an issue with nodemon compilation in ts-node 4.0 it is still slower. I will raise issues with them and let them fix it.

PR #269, opened 2017-12-13
ROC-2703: Unify claim response model

radoslawn-kainos:
### JIRA link

### Change description

Removed redundant response model, factored FullDefenceResponse out into a separate subtype

### Work checklist

- [ ] Yes
- [x] No


radoslawn-kainos:
I did a few things here differently than it's done elsewhere:
- used interfaces instead of classes
- made deserialization functions "static"
- main response deserializer delegates to appropriate variant depending on `responseType`

I think that using interfaces makes things much simpler (less boilerplate, less data shovelling required, as it's just a description of structure of plain objects rather than an encapsulated abstraction). It could be made even simpler, perhaps not even needing any deserialization (as we don't validate anything there), if all types used inside were interfaces (or string-based enums) as well.

I felt like doing this differently because we're going to extract the claim model into a separate package, so it might be a good idea to clean it up as much as possible.

PR #282, opened 2017-12-18
WIP - ROC-2703: Save part admission response

radoslawn-kainos:
### JIRA link

### Change description

Add part admission response model, and change to save it to claim store instead of redirecting to a handoff page.

### Work checklist

- [ ] Yes
- [ ] No
- [x] Not sure


src/test/features/response/routes/check-and-send.ts
```diff
@@ -191,20 +190,6 @@ describe('Defendant response: check and send page', () => {
                 .toLocation(ResponsePaths.counterClaimPage.evaluateUri({ externalId: claimStoreServiceMock.sampleClaimObj.externalId })))
           })
 
-          it('should redirect to partial-admission handoff page when defendant response is part admission', async () => {
```
yarpo: What happens now?

src/test/features/response/routes/partial-admission.ts
```diff
@@ -1,59 +0,0 @@
...
-import * as draftStoreServiceMock from '../../../http-mocks/draft-store'
-import * as claimStoreServiceMock from '../../../http-mocks/claim-store'
-
-import { checkCountyCourtJudgmentRequestedGuard } from './checks/ccj-requested-check'
-import { checkNotDefendantInCaseGuard } from './checks/not-defendant-in-case-check'
-
-const cookieName: string = config.get<string>('session.cookieName')
-const pagePath = ResponsePaths.partialAdmissionPage.evaluateUri({ externalId: claimStoreServiceMock.sampleClaimObj.externalId })
-
-describe('Defendant response: partial admission page', () => {
```
yarpo: why is this removed and no other tests added?

yarpo:
it looks good. My only concern is the lack of unit tests.

Yes, I know we probably had very low code coverage before, but can you double check if the new approach cannot be better unit tested? Thanks in advance :)

src/test/features/response/routes/check-and-send.ts
radoslawn-kainos: I guess it falls under "when everything is fine" now :wink:

Jokes aside, that's something I'd like to add. But the way these tests are structured, it doesn't feel like there's a good place for it, because there is no explicit test for defence response (when not counter-claiming) either.

src/test/data/draft/response/partAdmission.ts
```diff
@@ -0,0 +1,29 @@
+import { individualDetails } from '../partyDetails'
+
+export const responseDraft = {
```
geek-soft: We could call it draft as it will be the only one in the file. I do not feel strongly, it feels like you followed naming started in other place.

src/test/data/entity/response/partAdmission.ts
```diff
@@ -0,0 +1,11 @@
+import { individual } from '../party'
+
+export const responseData = {
+  defendant: individual,
+  moreTimeNeeded: 'no',
+  freeMediation: 'no',
+  responseType: 'PART_ADMISSION',
```
geek-soft: Would be great to have these on top but I do not feel strong, just my thing.

src/test/features/response/routes/check-and-send.ts
geek-soft: We should have the test documenting the behaviour of when user submits partial defence. The expected outcome is to take user to the confirmation page.

When test case is missing feel free to add it either now as this will be just few lines to separate PR.

src/test/features/response/routes/partial-admission.ts
geek-soft: I see now this is because `ResponsePaths.partialAdmissionPage` handoff page is removed and won't be used anymore. The confirmation page is reused now when response is submitted to claim-store. Only test case we need to add is in `check-and-send.ts` test file to reflect new redirect behaviour.

src/main/app/claims/responseModelConverter.ts
```diff
@@ -10,86 +13,212 @@ import { Company } from 'app/claims/models/details/yours/company'
...
+  switch (draft.option) {
+    case RejectPartOfClaimOption.AMOUNT_TOO_HIGH:
+      return PartAdmissionType.AMOUNT_TOO_HIGH
+    case RejectPartOfClaimOption.PAID_WHAT_BELIEVED_WAS_OWED:
+      return PartAdmissionType.PAID_WHAT_BELIEVED_WAS_OWED
+  }
+}
+
+function convertHowMuchOwed (draft: DraftHowMuchOwed): HowMuchOwed {
+  if (validator.validateSync(draft).length > 0) {
```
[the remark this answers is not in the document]
radoslawn-kainos: I added it because draft deserialization always put it there, even when it was not in the object to deserialize. But I just realised it's a required property and I can assume the values are always going to be there, so I will remove it.

src/test/data/draft/response/partAdmission.ts
radoslawn-kainos: What would you say about making it a default export? My thinking is, if a name is irrelevant, it might be better not to have it at all than have an arbitrary one.

src/main/app/claims/responseModelConverter.ts
```diff
@@ -10,86 +13,212 @@ import { Company } from 'app/claims/models/details/yours/company'
...
-    }
-    party.name = defendant.partyDetails.name
-    if (defendant.email) {
-      party.email = defendant.email.address
+function convertPartAdmission (draft: ResponseDraft): PartAdmissionResponse {
+  return {
+    ...convertCommon(draft),
+    responseType: ResponseType.PART_ADMISSION,
+    partAdmissionType: convertPartAdmissionType(draft.rejectPartOfClaim),
+    howMuchOwed: convertHowMuchOwed(draft.howMuchIsPaid),
```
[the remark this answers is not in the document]
radoslawn-kainos: Well spotted. I just expanded the unit test for the converter to catch things like this. Will update the PR shortly.

src/main/app/claims/responseModelConverter.ts
```diff
@@ -10,86 +13,212 @@ import { Company } from 'app/claims/models/details/yours/company'
...
+  return draft.getPopulatedRowsOnly()
+  .map(row => ({
+    type: mapEvidenceType(row.type),
+    description: row.description
+  }))
+
+  function mapEvidenceType (draft: DraftEvidenceType): EvidenceType {
+    switch (draft.value) {
+      case DraftEvidenceType.CONTRACTS_AND_AGREEMENTS.value:
+        return EvidenceType.CONTRACTS_AND_AGREEMENTS
```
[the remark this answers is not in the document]
radoslawn-kainos: Not sure if I understand what you mean here. I'd prefer EvidenceType to be unaware of its draft counterpart, and assuming that both use exact same values might be dangerous. If the values are assumed to be exactly the same, why use two separate types for it?

src/test/features/response/routes/check-and-send.ts
radoslawn-kainos: Updated PR with a new test added

radoslawn-kainos:
Created another PR - #286 - with the branch name corresponding to the claim-store one.

PR #286, opened 2017-12-20
ROC-2703: Save part admission response

radoslawn-kainos:
### JIRA link

### Change description

Add part admission response model, and change to save it to claim store instead of redirecting to a handoff page.

### Work checklist

- [ ] Yes
- [ ] No
- [x] Not sure


src/main/app/claims/responseModelConverter.ts
```diff
@@ -10,86 +13,205 @@ import { Company } from 'app/claims/models/details/yours/company'
...
+  throw new Error(`Unsupported defence type: ${draft.option}`)
+}
+
+function convertPartAdmissionType (draft: RejectPartOfClaim): PartAdmissionType {
+  switch (draft.option) {
+    case RejectPartOfClaimOption.AMOUNT_TOO_HIGH:
+      return PartAdmissionType.AMOUNT_TOO_HIGH
+    case RejectPartOfClaimOption.PAID_WHAT_BELIEVED_WAS_OWED:
+      return PartAdmissionType.PAID_WHAT_BELIEVED_WAS_OWED
+  }
```
geek-soft: Shouldn't we throw an error when conversion failed? In theory it looks like very unlikely scenario because task list should not allow submitting response without making sure draft is valid. It is more to fail fast before making call to backend which would return 400.

src/main/app/claims/responseModelConverter.ts
```diff
@@ -10,86 +13,205 @@ import { Company } from 'app/claims/models/details/yours/company'
...
+function convertHowMuchOwed (draft: DraftHowMuchOwed): HowMuchOwed {
+  return {
+    amount: draft.amount,
+    explanation: draft.text
+  }
+}
+
+function convertTimeline (draft: DraftTimeline): Timeline {
+  return draft.getPopulatedRowsOnly()
+  .map(row => ({
```
geek-soft: That mapping looks like pointless, we could just return what `draft.getPopulatedRowsOnly()` returns.

src/main/app/claims/responseModelConverter.ts
```diff
@@ -10,86 +13,205 @@ import { Company } from 'app/claims/models/details/yours/company'
...
+      case DraftEvidenceType.OTHER.value:
+        return EvidenceType.OTHER
     }
-    if (defendant.mobilePhone) {
-      party.mobilePhone = defendant.mobilePhone.number
+  }
+}
+
+function convertPaymentPlan (draft: DefendantPaymentPlan): PaymentPlan {
+  if (validator.validateSync(draft).length > 0) {
```
geek-soft: I don't like the fact validator is used to make conversion. Converter should take what is given and validate it without any additional check. Alternatively we can build contract in such a way that converter validates object on the `fromDraft` method and throws an exception when invalid.

src/test/app/claims/responseModelConverter.ts
```diff
@@ -40,7 +42,14 @@ describe('ResponseModelConverter', () => {
       const responseDraft = prepareResponseDraft(partyDetails)
       const responseData = prepareResponseData(party)
 
-      expect(ResponseModelConverter.convert(responseDraft)).to.deep.equal(responseData)
+      expect(ResponseModelConverter.fromDraft(responseDraft)).to.deep.equal(responseData)
     })
   })
+
+  it('converts part admission response', () => {
```
geek-soft: We should at least test two main types of part admission. That test is pretty optimistic and conversion proved to be difficult as we had production issue due to incorrectly implemented conversion.

src/test/data/draft/response/partAdmission.ts
```diff
@@ -0,0 +1,40 @@
...
+    },
+    email: {
+      address: 'individual@example.com'
+    }
+  },
+  moreTimeNeeded: {
+    option: 'no'
+  },
+  freeMediation: {
+    option: 'no'
```
geek-soft: Would be good to test the opposite option as well.

src/test/data/draft/response/partAdmission.ts
```diff
@@ -0,0 +1,40 @@
...
+  },
+  response: {
+    type: {
+      value: 'PART_ADMISSION'
+    }
+  },
+  rejectPartOfClaim: {
+    option: 'amountTooHigh'
```
geek-soft: We need to test part admission due to amount is believed to be already paid.

src/test/data/draft/response/partAdmission.ts
```diff
@@ -0,0 +1,40 @@
...
+  },
+  howMuchOwed: {
+    amount: 42,
+    text: 'reasons'
+  },
+  impactOfDispute: {
+    text: 'very much'
+  },
+  timeline: [],
```
geek-soft: We need to test timeline and evidence conversion as well.

geek-soft:
Generally frontend conversion for part admission due to amount owed being paid is missing the same as in the backend.

PR #291, opened 2018-01-04
ROC-2855-Refactor-claim-status-logic

jacmurphy:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-2855

### Change description
Refactoring claim status to make it more easily testable using typescript.

### Updates in other projects
https://github.com/hmcts/cmc-common-frontend/pull/81

- [ ] Yes
- [X ] No


src/test/app/claims/models/claim.ts
```diff
@@ -148,4 +154,73 @@ describe('Claim', () => {
       expect(claim.defendantOffer).to.be.eq(undefined)
     })
   })
+
+  describe('claimStatus', () => {
+
+    it('should return ELIGIBLE_FOR_CCJ if defendant has not responded before the deadline', () => {
+      const claim = buildClaim()
```
rlewan: You can use a `beforeEach` block to build the claim. This way you can avoid doing the same initialisation in each test. For example:
```
describe('claimStatus', () => {
  let claim

  beforeEach(() => {
    claim = buildClaim()
  })
```

src/test/app/claims/models/claim.ts
```diff
@@ -148,4 +154,73 @@ describe('Claim', () => {
...
+      claim.response = {
+        responseType: ResponseType.FULL_DEFENCE,
+        defenceType: DefenceType.DISPUTE,
+        defence: 'defence reasoning',
+        freeMediation: FreeMediationOption.NO,
+        defendant: new Individual().deserialize(individual)
+      }
+      expect(claim.status).to.be.equal(ClaimStatus.CLAIM_REJECTED)
+    })
```
rlewan: It would be nice if we had a test case for `NO_RESPONSE` as well, just to be sure how this component behaves in all possible scenarios.

src/main/app/claims/models/claim.ts
```diff
@@ -111,4 +113,22 @@ export class Claim {
 
     return !this.countyCourtJudgmentRequestedAt && this.remainingDays < 0 && !this.respondedAt
   }
+
+  get status (): ClaimStatus {
+    if (this.countyCourtJudgmentRequestedAt) {
+      return ClaimStatus.CCJ_REQUESTED
+    } else if (this.eligibleForCCJ) {
+      return ClaimStatus.ELIGIBLE_FOR_CCJ
+    } else if (toBoolean(config.get<boolean>('featureToggles.offer')) && this.settlement && this.response.responseType === ResponseType.FULL_DEFENCE) {
```
rlewan: You can use `FeatureToggles.isEnabled('offer')` instead. See `featureToggle.ts`.

src/test/app/claims/models/claim.ts
jacmurphy: NO_RESPONSE test added now 👍

src/test/app/claims/models/claim.ts
```diff
@@ -148,4 +154,78 @@ describe('Claim', () => {
       expect(claim.defendantOffer).to.be.eq(undefined)
     })
   })
+
+  describe('claimStatus', () => {
```
yarpo: Please add more cases, eg. more time requested + eligible for ccj -> make sure that when claim matches both, "the more important" status is returned.

jacmurphy: Done 👍

src/main/app/claims/models/claim.ts
```diff
@@ -88,4 +92,44 @@ export class Claim {
...
+    } else if (this.moreTimeRequested) {
+      return ClaimStatus.MORE_TIME_REQUESTED
+    } else if (!this.response) {
+      return ClaimStatus.NO_RESPONSE
+    } else {
+      throw new Error('Unknown Status')
+    }
+  }
+
+  private isFreeMediationRequested () {
```
rlewan: Good idea with extracting logical expressions into methods 👍

PR #297, opened 2018-01-09
Move clearing dropdown to last moment

timja-kainos:
Seems to stop the duplicate address options problem


timja-kainos:
Not sure how unit tests would catch this one, seems to be a timing issue

PR #298, opened 2018-01-09
ROC-2005 Fix postcode validation

timja-kainos:
This ensures that the validation isn't skipped if a user enters a name and a postcode and does nothing else


src/main/app/forms/models/address.ts
```diff
@@ -62,7 +63,10 @@ export class Address implements CompletableTask {
   postcode?: string
 
   @ValidateIf(o => !o.addressVisible && !o.addressSelectorVisible, { groups: ['claimant', 'defendant', 'response'] })
-  @IsNotBlank({ message: ValidationErrors.POSTCODE_REQUIRED, groups: ['claimant', 'defendant', 'response'] })
+  @AllFieldsArePopulated(['postcode', 'line1'], {
```
geek-soft: These extra validation rules required by postcode lookup are so unreadable :(
  
It is so complex that at least we should have proper unit tests to describe the rules if we want to accept that tech debt.

geek-soft: Why didn't you include 'city' property as well? I cannot find reason for having just two of three required.

src/main/app/forms/validation/validators/allFieldsArePopulated.ts
```diff
@@ -0,0 +1,29 @@
...
+  ValidationArguments,
+  ValidationOptions,
+  ValidatorConstraint,
+  ValidatorConstraintInterface
+} from 'class-validator'
+
+@ValidatorConstraint()
+export class AllFieldsArePopulatedConstraint implements ValidatorConstraintInterface {
+  validate (value: any, args?: ValidationArguments) {
+    return args.constraints.every(field => args.object[field] && !!args.object[field].trim())
```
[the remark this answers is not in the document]
timja-kainos: its not technically needed, as line 1 will always be undefined when it is hidden, I added the trim to make it more generic.
Do you think I should remove it?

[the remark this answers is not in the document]
geek-soft: Yes. Removing it would also make the documentation more accurate:

> Verifies if all of the fields in given object are populated with "truthy" value.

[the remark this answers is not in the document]
rlewan: I think I'd extract the body of this function into another function which checks if the argument is a string and only then calls `trim`.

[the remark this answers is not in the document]
timja-kainos: will go with @geek-soft 's approach, its not needed now and can be added later if needed

src/main/app/forms/models/address.ts
rlewan: I agree that it's quite unintuitive as it is now. What do you think about renaming the annotation? `@ExtraFormFieldsArePopulated`? `@SupplementaryFormFieldsArePopulated`?

src/main/app/forms/validation/validators/allFieldsArePopulated.ts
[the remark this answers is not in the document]
rlewan: I mean the arrow function body.

src/test/app/forms/models/address.ts
```diff
@@ -148,6 +148,35 @@ describe('Address/CorrespondenceAddress', () => {
...
+        it('should reject when postcode fields are not populated', () => {
+          const address = new ClassFunction()
+          address.addressVisible = false
+          address.addressSelectorVisible = false
+          const errors = validator.validateSync(address)
+
+          expect(errors.length).to.equal(1)
+        })
+      })
+      context('addressVisible is false and addressSelectorVisible is true', () => {
```
yarpo: empty line before and after

src/test/app/forms/models/address.ts
```diff
@@ -148,6 +148,35 @@ describe('Address/CorrespondenceAddress', () => {
...
+        it('should reject when address is not selected', () => {
+          const address = new ClassFunction()
+          address.addressVisible = false
+          address.addressSelectorVisible = true
+          const errors = validator.validateSync(address)
+          expect(errors.length).to.equal(1)
+        })
+      })
+      context('addressVisible is true and addressSelectorVisible is true', () => {
```
yarpo: same as above

src/test/app/forms/models/address.ts
```diff
@@ -148,6 +148,37 @@ describe('Address/CorrespondenceAddress', () => {
 
         expect(errors.length).to.equal(0)
       })
+
+      context('addressVisible is false and addressSelectorVisible is false', () => {
```
geek-soft: I would love more plain English version describing how the scenario triggered e.g. `when postcode lookup has not been triggered` or something.

Would be also good to have assertion on the expected message, nut just the number of errors.

src/main/app/forms/validation/validators/extraFormFieldsArePopulated.ts
```diff
@@ -0,0 +1,29 @@
...
+export class ExtraFormFieldsArePopulatedConstraint implements ValidatorConstraintInterface {
+  validate (value: any, args?: ValidationArguments) {
+    return args.constraints.every(field => !!args.object[field])
+  }
+}
+
+/**
+ * Verifies if all of the fields in given object are populated with "truthy" value.
+ */
+export function ExtraFormFieldsArePopulated (fields: string[], validationOptions?: ValidationOptions) {
```
geek-soft: What about `FieldsArePopulated`? I personally do not understand this `extra form` bit. Why extra? Form bit is also unnecessary as validation should usable anywhere, not just for the form data.

src/main/app/forms/models/address.ts
```diff
@@ -62,7 +63,10 @@ export class Address implements CompletableTask {
   postcode?: string
 
   @ValidateIf(o => !o.addressVisible && !o.addressSelectorVisible, { groups: ['claimant', 'defendant', 'response'] })
-  @IsNotBlank({ message: ValidationErrors.POSTCODE_REQUIRED, groups: ['claimant', 'defendant', 'response'] })
+  @ExtraFormFieldsArePopulated(['postcode', 'postcodeLookup'], {
```
geek-soft: That makes a bit more sense to me ;)

PR #311, opened 2018-01-16
ROC-2990: Remove validation from phone no and make it optional

sabahirfan:
### JIRA link
[ROC-2990](https://tools.hmcts.net/jira/browse/ROC-2990)

### Change description
This PR removes the mobile no validation from the phone and makes it optional. Makes it as free text up to 30 characters allowed. 

### Work checklist

- [ ] Yes
- [x] No


src/main/app/forms/models/mobilePhone.ts
```diff
@@ -29,7 +26,4 @@ export class MobilePhone implements CompletableTask {
     return this
   }
 
-  isCompleted (): boolean {
-    return !!this.number && this.number.length > 0
```
timja-kainos: why remove this? I assume its still mandatory to provide a number?

src/main/features/claim/views/claimant-mobile.njk
```diff
@@ -12,7 +12,7 @@
         <form method="post">
           {{ csrfProtection(csrf) }}
 
-          {{ textInput('Mobile number', 'number', form) }}
+          {{ textInput('', 'number', form) }}
```
timja-kainos: shouldn't there be a visually hidden label here? @radoslawn-kainos

radoslawn-kainos: Yup, can't leave the field with an empty label. Setting [`labelClass`](https://github.com/hmcts/cmc-common-frontend/blob/master/macros/form.njk#L1) param to `visually-hidden` should be enough. Might require modifying `textInput` macro to pass that on.

sabahirfan: done

PR #313, opened 2018-01-17
ROC-3008: link defendant with claim on first login

yarpo:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-3008

### Change description
When user (defendant) go through first contact, clicks link "activate account", sets password, but do not log in.

Then on first login attempt the process of linking up claim with user will be finished off.

### Work checklist

- [ ] ~Unit tests added where applicable~
- [ ] ~Route tests added for new pages~
- [ ] ~New pages included in a11y tests~
- [ ] ~Task list and task completeness checks updated~
- [ ] ~Check and send page updated~
- [ ] ~Routes, page content and page flows of new features are toggled~
- [ ] ~UI changes look good on mobile~
- [ ] ~Required Google Analytics events are being sent~

- [x] Yes
- [ ] No


src/main/routes/receiver.ts
```diff
@@ -139,11 +139,12 @@ export default express.Router()
                                res: express.Response,
                                next: express.NextFunction): Promise<void> => {
       const cookies = new Cookies(req, res)
+      let user
 
       try {
         const authenticationToken = await getAuthToken(req)
         if (authenticationToken) {
-          const user = await IdamClient.retrieveUserFor(authenticationToken)
+          user = await IdamClient.retrieveUserFor(authenticationToken)
```
timja-kainos: do you plan to remove `AppPaths.linkDefendantReceiver.uri` in a separate PR?
as that was one of the goals of this

yarpo: I'd rather do it separately. 

What's more we will have to keep part of it (at least to redirect) as this url is stored in db:
image

in `continue_url` column, idam db.

rlewan: Also I think we can keep the `const` here, I can't see any reassignments in the remaining part of this call handler.

src/main/routes/receiver.ts
```diff
@@ -157,11 +158,18 @@ export default express.Router()
           // re-set state cookie as it was cleared above, we need it in this case
           cookies.set(stateCookieName, req.query.state, { sameSite: 'lax' })
           return res.redirect(FirstContactPaths.claimSummaryPage.uri)
+        } else {
+          const letterHolderId: string = getLetterHolderId(req, user)
```
rlewan: Can there be multiple unlinked cases?

src/main/routes/receiver.ts
timja-kainos: @yarpo hmm, I think we could get away without it, or if we think we need it just add a redirect that to the new url making sure the query string parameters are preserved

src/main/routes/receiver.ts
```diff
@@ -157,11 +158,18 @@ export default express.Router()
...
+          if (letterHolderId && user.isInRoles(`letter-${letterHolderId}`)) {
+            const claim: Claim = await ClaimStoreClient.retrieveByLetterHolderId(letterHolderId)
+            logger.debug(`Linking user ${user.id} claim ${claim.id}`)
```
timja-kainos: How about `Linking user ${user.id} to claim ${claim.id}`

src/main/routes/receiver.ts
timja-kainos: yes, was just about to point this out, the `getLetterHolderId` function cheats because it knows that on registration there will only ever be 1.

src/main/routes/receiver.ts
yarpo: @rlewan we cannot :(
image

timja-kainos: @yarpo Why do you need the user extracted at that level?

src/main/routes/receiver.ts
yarpo: You're right. We should cover it as well, even if it's very unlikely to happen.

src/main/routes/receiver.ts
yarpo: @timja-kainos yes, as I need it here:

```
const letterHolderId: string = getLetterHolderId(req, user)
          if (letterHolderId && user.isInRoles(`letter-${letterHolderId}`)) {
            const claim: Claim = await ClaimStoreClient.retrieveByLetterHolderId(letterHolderId)
            logger.debug(`Linking user ${user.id} claim ${claim.id}`)
            if (!claim.defendantId) {
              await ClaimStoreClient.linkDefendant(claim.id, user.id)
            }
          }
```

src/main/routes/receiver.ts
```diff
@@ -157,11 +136,19 @@ export default express.Router()
           // re-set state cookie as it was cleared above, we need it in this case
           cookies.set(stateCookieName, req.query.state, { sameSite: 'lax' })
           return res.redirect(FirstContactPaths.claimSummaryPage.uri)
+        } else {
+          (user as User).getLetterHolderIdList().forEach(async (letterHolderId) => {
```
timja-kainos: why cast here?

Bakaloy (approved):

PR #324, opened 2018-01-18
Set node environment in CNP

rlewan:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-2914


infrastructure/main.tf
```diff
@@ -39,7 +39,7 @@ module "citizen-frontend" {
     WEBSITE_NODE_DEFAULT_VERSION = "8.9.0"
 
     // Node specific vars
-    NODE_ENV = "${var.env}"
+    NODE_ENV = "${contains(list("dev", "demo", "test", "prod"), var.env) ? var.env : "dev"}"
```
timja-kainos: I would probably just go for a `== prod` here not sure if we get any benefit from the others, considering they probably won't ever be set to those values, at least for test

rlewan: Makes sense 👍

PR #329, opened 2018-01-23
ROC-3020: Add auth header for link status backend call

sabahirfan:
### JIRA link
[ROC-3020](https://tools.hmcts.net/jira/browse/ROC-3020)

### Change description
This PR adds auth header for defendant link status backend call.


timja-kainos:
There's no user there this change is impossible

PR #333, opened 2018-01-24
ROC-3135: Provide classlist polyfill

rlewan:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-3135

### Change description

Provides a polyfill to make postcode lookup client code work on IE9.

- [ ] Yes
- [X] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "build-url": "^1.0.9",
     "class-transformer": "^0.1.7",
     "class-validator": "^0.7.0",
+    "classlist-polyfill": "^1.2.0",
     "config": "^1.24.0",
     "continuation-local-storage": "^3.2.1",
     "cookie-parser": "^1.3.3",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cmc-citizen-frontend",
-  "version": "1.2.0",
+  "version": "1.1.1",
   "private": true,
   "engines": {
     "node": ">=8.0.0"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cmc-citizen-frontend",
-  "version": "1.1.1",
+  "version": "1.2.0",
   "private": true,
   "engines": {
     "node": ">=8.0.0"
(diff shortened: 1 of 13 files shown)


src/main/features/claim/views/claimant-company-details.njk
```diff
@@ -7,6 +7,7 @@
 
 {% block additional_scripts %}
   <script src="{{ asset_paths['js_vendor'] }}/nodelist-foreach-polyfill.js"></script>
+  <script src="{{ asset_paths['js_vendor'] }}/classlist-polyfill.js"></script>
```
geek-soft: It would be better to have one macro that renders all script tags necessary for postcode lookup (including postcode lookup JS itself). If something needs to be added / changed then we have single place to carry on that change.

rlewan: @geek-soft do you mean to have a macro that would just render the script tags and include that macro on pages where it's needed? Or make postcode lookup somehow include those by itself? I'm not sure if latter would be possible as I think the pattern to follow is to include script tags at the bottom of the DOM tree.

radoslawn-kainos: @geek-soft One possible problem with that approach is, what if some script was needed for more than one reason? I think, either use something that manages dependencies properly, or keep it dumb.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "cmc-citizen-frontend",
-  "version": "1.1.1",
+  "version": "1.2.0",
```
timja-kainos: why are you doing this?

src/main/features/claim/views/claimant-company-details.njk
geek-soft: @radoslawn-kainos that is a valid point.

The intermediate solution for now could be including polyfills globally for any service page.

The more strategic solution could be using dependency manager like webpack to bundle polyfills at the build time into the bundle.

Lets have a chat about this tomorrow to see how we can improve it.

package.json
rlewan: I bumped because we have new code that will be deployed during next release and incrementing the feature version made more sense to me. It's probably unrelated to this particular PR, I can remove this.

timja-kainos: please remove, its unrelated

PR #334, opened 2018-01-24
ROC-3132 Fix custom event sender in IE

timja-kainos:
Because of the `for...of` loop, IE didn't work.
Reworked the code so the loop isn't needed
It assumes the text will always be the first element, have tested on
eligibility

### DIFF
--- a/src/main/public/js/analytics-custom-event-sender.js
+++ b/src/main/public/js/analytics-custom-event-sender.js
@@ -30,13 +30,15 @@ $(function () {
       throw new Error('Label element is required')
     }
 
-    for (var content of labelElement.contents()) {
-      const text = $(content).text().trim()
-      if (text && content.class !== 'form-hint') {
-        return text
-      }
-    }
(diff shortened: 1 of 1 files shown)


src/main/public/js/analytics-custom-event-sender.js
```diff
@@ -30,13 +30,7 @@ $(function () {
...
     }
 
-    for (var content of labelElement.contents()) {
-      const text = $(content).text().trim()
-      if (text && content.class !== 'form-hint') {
-        return text
-      }
-    }
-    return undefined
+    return labelElement.contents()[0].textContent.trim()
```
yarpo: are we sure this will always complete successfully?

timja-kainos: nope, needs a bit more testing really

timja-kainos: it works for eligibility

PR #293, opened 2018-01-08
ROC-2005: Postcode lookup (alternative implementation)

geek-soft:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-2005

### Change description

An alternative implementation of postcode lookup that does not require changes in the data model which helps to keep postcode lookup in progressive enhancement spirit.

The validation messages are amended by client side JS code to sync it with widgets state.

PR require unreleased version of address macro. See PR: https://github.com/hmcts/cmc-common-frontend/pull/80

### Work checklist

- [ ] Unit tests added where applicable
- [ ] Route tests added for new pages
- [ ] New pages included in a11y tests
- [ ] Task list and task completeness checks updated
- [ ] Check and send page updated
- [ ] Routes, page content and page flows of new features are toggled 
- [x] UI changes look good on mobile
- [ ] Required Google Analytics events are being sent 

### Developer self-QA run statement

- [x] I have clicked through the running application to see if all changes I made actually work.

### Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #338, opened 2018-01-25
ROC-2016 Deserialize amount having comma

dharmendrak:


src/main/features/claim/form/models/claimAmountRow.ts
```diff
@@ -38,13 +38,27 @@ export class ClaimAmountRow {
     return new ClaimAmountRow(undefined, undefined)
   }
 
+  static translateToNumber (input?: any): any {
+    const numberValue = _.toNumber(input)
+    const isExponential = numberValue.toString().match(new RegExp('\\d*[Ee][+-]?\\d*?'))
```
rlewan: Did the UX say we need this? What sane person would use scientific notation when issuing a money claim?

src/main/features/claim/form/models/claimAmountRow.ts
```diff
@@ -38,13 +38,27 @@ export class ClaimAmountRow {
     return new ClaimAmountRow(undefined, undefined)
   }
 
+  static translateToNumber (input?: any): any {
```
rlewan: Can we use more precise types? I would expect `(input?: string): number`.

src/main/features/claim/form/models/claimAmountRow.ts
```diff
@@ -38,13 +38,27 @@ export class ClaimAmountRow {
...
+
   static fromObject (value?: any): ClaimAmountRow {
     if (!value) {
       return value
     }
 
     const reason = value.reason || undefined
-    const amount = value.amount ? _.toNumber(value.amount) : undefined
+    const amount = value.amount
+      ? ClaimAmountRow.translateToNumber(value.amount.replace(new RegExp(/,/g), ''))
```
rlewan: I think it's not necessary to do `new RegExp`, `/,/` literal should be enough.

src/main/features/claim/form/models/claimAmountRow.ts
```diff
@@ -38,13 +38,27 @@ export class ClaimAmountRow {
     return new ClaimAmountRow(undefined, undefined)
   }
 
+  static translateToNumber (input?: any): any {
+    const numberValue = _.toNumber(input)
+    const isExponential = numberValue.toString().match(new RegExp('\\d*[Ee][+-]?\\d*?'))
+
+    if (!isNaN(numberValue) && !isExponential) {
+      return numberValue
```
timja-kainos: I assume this will remove comma's that the url has entered?
Probably ok, but just a thought

geek-soft:
@dharmendrak please provide PR description.

src/main/features/claim/form/models/claimAmountRow.ts
geek-soft: There is already `toNumberOrUndefined` function which looks like better place to put that change if I understand this correctly that it is about handling comma thousand separators in UK localised systems.

Fixing problem in just a single place is very short sighted solution.

src/main/features/claim/form/models/claimAmountRow.ts
```diff
@@ -38,13 +38,27 @@ export class ClaimAmountRow {
...
+    } else {
+      return input
```
geek-soft: Method should return undefined. A `toNumberOrUndefined` does it right.

src/main/features/claim/form/models/claimAmountRow.ts
geek-soft: @dharmendrak I do not see a reason why removing coma is done in one place and while conversion to numeric type is done in another place. Would you expect parsing function in Java to accept pre-processed value? Please keep both things together, the goal if I understand correctly is to handle numbers with coma in deserialisation.

src/main/features/claim/form/models/claimAmountRow.ts
dharmendrak: citizen frontend had amount restricted to 10,000 however user still can enter number more than 13 digit, when it get changed to exponential. and our validation has to handle exponential number and it looses it's value during validation process

src/main/features/claim/form/models/claimAmountRow.ts
dharmendrak: return type can be a number or string (original value)

src/main/features/claim/form/models/claimAmountRow.ts
dharmendrak: @geek-soft Can interest rate be entered more than three digits.
@rlewan this replace remove all , in the number, I mean it can be more than one.

src/main/features/claim/form/models/claimAmountRow.ts
dharmendrak: method returns a number or string (original value) so that user input is not lost in conversion as the requirement

src/main/features/claim/form/models/claimAmountRow.ts
dharmendrak: else block return's original value

src/main/common/utils/numericUtils.ts
```diff
@@ -14,3 +14,37 @@ export function toNumberOrUndefined (value: any): number {
 
   return isNaN(numberVal) ? undefined : numberVal
 }
+
+export function containsAValidComma (input: string): boolean {
```
geek-soft: Please do not export function. It does not look like something reusable outside.

src/main/common/utils/numericUtils.ts
```diff
@@ -14,3 +14,37 @@ export function toNumberOrUndefined (value: any): number {
...
+  const numberValue = parseFloat(output)
+
+  if (!output.includes(',') && !isNaN(numberValue) && !this.isExponential(numberValue)) {
+    return numberValue
+  } else {
+    return output
+  }
+}
+
+export function isExponential (numberValue: number) {
```
geek-soft: Do not export that function unless needed. If you exposing it make it well tested.

src/main/common/utils/numericUtils.ts
```diff
@@ -14,3 +14,37 @@ export function toNumberOrUndefined (value: any): number {
...
   return isNaN(numberVal) ? undefined : numberVal
 }
+
+export function containsAValidComma (input: string): boolean {
+  const output: string[] = input.split(',')
+
+  return output[output.length - 1].length >= 3
+}
+
+export function toAValidFloatOrOriginalValue (input: string): any {
```
geek-soft: Looking at the method I would say it we do not have to distinguish numbers with fraction from numbers without it.

geek-soft: Another thing is that this method is doing two things, it approaches deserialisation and if impossible it gives up by returning original. That is a bot too much IMO.

I propose the following approach:
1. deserialisation logic is responsible for deserialisation only and if impossible returns NaN or undefined
2. UI component is responsible for rendering correct value in the following way:
2.1. if value is valid it renders it as is using `valueFor`
2.2. if value is invalid the raw value is rendered from the form using `rawDataFor`

See: https://github.com/hmcts/cmc-citizen-frontend/pull/293/files#diff-8e8267522c3e829b11e0fcae1908cdb4R50 - implementation that stores raw form data in form object

src/main/common/utils/numericUtils.ts
rlewan: Also please please refer to this as a "thousands separator". "valid comma" sounds a little confusing and does not carry much information.

src/main/common/utils/numericUtils.ts
```diff
@@ -14,3 +14,37 @@ export function toNumberOrUndefined (value: any): number {
...
+  }
+}
+
+export function isExponential (numberValue: number) {
+  return numberValue.toString().match(new RegExp('\\d*[Ee][+-]?\\d*?'))
+}
+
+export function toAValidNumberOrOriginalValue (input: string): any {
+
+  const output: string = this.containsAValidComma(input) ? input.replace(new RegExp(/,/g), '') : input
```
rlewan: I think you can just use the regular expression literal `/,/` instead of `new RegExp(/,/g)`.

src/main/common/utils/numericUtils.ts
rlewan: Also why do we need to handle scientific notation? I would just discard it as invalid value. No sane person would use scientific notation when claiming for money.

src/main/common/utils/numericUtils.ts
rlewan: I just check and the `g` switch is needed, so `/,/g` would suffice (no need to call `new RegExp`).

src/main/common/utils/numericUtils.ts
```diff
@@ -14,3 +14,8 @@ export function toNumberOrUndefined (value: any): number {
 
   return isNaN(numberVal) ? undefined : numberVal
 }
+
+export function containsAThousandSeparator (input: string): boolean {
+  const output: string[] = input.split(',')
+  return output[output.length - 1].length >= 3
```
rlewan: This only checks the last part of the number string. What about e.g. `1,00,00,000`? That's not a correctly formatted number but will return true.

rlewan: And for `1,000000000` it returns true as well. I think that if thousands separator is used then it consistent formatting in all places should be enforced.

dharmendrak: numerals handles 1,0000000, however if any comma is present in less than 3, it is wrong value

dharmendrak:
jenkin issue

PR #350, opened 2018-01-29
ROC-2935: Filter out illegal characters from user input

radoslawn-kainos:
### JIRA link

[ROC-2935](https://tools.hmcts.net/jira/browse/ROC-2935)

### Change description

Filter out illegal characters from user input. Happens before form validation.

### Work checklist

- [ ] Yes
- [x] No

### DIFF
--- a/src/main/app/forms/validation/formValidator.ts
+++ b/src/main/app/forms/validation/formValidator.ts
@@ -2,6 +2,7 @@ import * as express from 'express'
 import { Validator, ValidationError } from 'class-validator'
 
 import { Form } from 'forms/form'
+import * as _ from 'lodash'
 
 type Constructor<T> = { new(): T }
 type Mapper<T> = (value: any) => T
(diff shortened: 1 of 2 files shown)


src/main/app/forms/validation/formValidator.ts
```diff
@@ -26,7 +27,7 @@ export class FormValidator {
     }
 
     return (req: express.Request, res: express.Response, next: express.NextFunction) => {
-      const model: T = modelTypeMapper(req.body)
+      const model: T = modelTypeMapper(removeIllegalCharacters(req.body))
```
timja-kainos: does everything go through the formValidator?

radoslawn-kainos: I can't say for sure, but do you think it would make sense to assume that everything, that is a form submission, should?

timja-kainos: seems safe to me

PR #354, opened 2018-01-31
ROC-3004: Additional link pages in footer

jacmurphy:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-3004

### Change description
Added Pages:
- Cookies
- Privacy
- Policy
- Terms and conditions
- Help using GOV.UK

- [ ] Yes
- [X] No

### I had to recreate a separate PR to fix an issue with Jenkins ###


src/main/routes/cookies.ts
```diff
@@ -0,0 +1,9 @@
+import * as express from 'express'
+
+import { Paths as AppPaths } from 'app/paths'
+
+/* tslint:disable:no-default-export */
+export default express.Router()
+  .get(AppPaths.cookiesPage.uri, function (req, res) {
```
geek-soft: These should be behind feature toggle as well.

Another things is that "fat arrow" syntax is preferred. Please also add types.

src/main/modules/nunjucks/index.ts
```diff
@@ -101,6 +103,10 @@ export class Nunjucks {
     nunjucksEnv.addGlobal('UnemploymentType', UnemploymentType)
     nunjucksEnv.addGlobal('BankAccountType', BankAccountType)
     nunjucksEnv.addGlobal('ClaimStatus', ClaimStatus)
+    nunjucksEnv.addGlobal('AppPaths', AppPaths)
+    if (FeatureToggles.isEnabled('finePrint')) {
+      nunjucksEnv.addGlobal('cookieText', `GOV.UK uses cookies make the site simpler. <a href="${AppPaths.cookiesPage.uri}">Find out more about cookies</a>`)
```
rlewan: Where is this global used? Also, will this text be translated?

src/main/views/cookies.njk
```diff
@@ -0,0 +1,213 @@
...
+        </tbody>
+      </table>
+
+      <h2 class="heading-small">{{ t('To identify you when you come back to the service') }}</h2>
+      <p>{{ t('We use authentication cookies to identify you when you return to the service.') }}</p>
+
+      <table class="form-group">
+        <thead>
+        <tr>
+          <th scope="col">Name</th>
```
rlewan: I think these headers need to be translated (here and in other places in this file).

src/main/modules/nunjucks/index.ts
robertparkinson: in layout.njk add t() at line 9: 
```<p>{{ t(cookieText) | safe }}</p>```

jacmurphy: Thanks for spotting this!

It is the seen_cookie_messag - This will be a dropdown pop up when a user goes to our service for the first time, or it is the first time they view this in the last month.

PR #355, opened 2018-01-31
ROC-2016 Deserialize number having comma

dharmendrak:


src/main/common/utils/numericUtils.ts
```diff
@@ -14,3 +14,17 @@ export function toNumberOrUndefined (value: any): number {
...
+  for (let i: number = 0; i < output.length; i++) {
+    if (i === output.length - 1 && output[i].length < 3) {
+      return false
+    } else if (output[i].length > 3) {
+      return false
+    }
+  }
+
+  return true
+}
```
rlewan: What do you think about using a regular expression to check this?
```
export function usesThousandsSeparators (input: string): boolean {
  return input.match(/^[1-9]\d{0,2}(\,\d{3})+(\.\d*)?$/)
}
```
It's a little more concise.

dharmendrak: better

src/main/common/utils/numericUtils.ts
```diff
@@ -14,3 +14,17 @@ export function toNumberOrUndefined (value: any): number {
 
   return isNaN(numberVal) ? undefined : numberVal
 }
+
+export function containsAThousandSeparator (input: string): boolean {
```
geek-soft: Please do not expose that function, no point in having it and then using it in tens of places. Change `toNumberOrUndefined` which is a central point of deserialising numbers.

A `toNumberOrUndefined` should use numeral as an improvement on top of dumb `_.toNumber(strVal)`

Plase do the following:
1. move `containsAThousandSeparator` to `common/utils/numericUtils.ts`, make it private and reuse within `toNumberOrUndefined` as an improvement to that method
2. Use `toNumberOrUndefined ` in all places that have:

```const amount = value.amount ? !value.amount.toString().includes(',') || containsAThousandSeparator(value.amount.toString())  ? numeral(value.amount).value() : value.amount : undefined```

3. Add tests that ensure deserialisation logic works fine with thousand separators, nobody is interested in knowing how `containsAThousandSeparator` works.

File: src/main/common/utils/numericUtils.ts [hunk 3]
```diff
@@ -10,7 +10,19 @@ export function toNumberOrUndefined (value: any): number {
...
+      : strVal
+    : undefined
+
+  // returns undefined for
+  return !numberVal ? undefined : numberVal
+}
+
+function containsAThousandSeparator (input: string): boolean {
+  const matchedArray = input.match(/^[1-9]\d{0,2}(\,\d{3})+(\.\d*)?$/)
+  return matchedArray ? matchedArray.length > 0 : false
```
rlewan: I didn't check the return value of `match` method, apologies. In this case I think we could do this to make the code simpler:
```typescript
const THOUSANDS_SEPARATORS_REGEXP = /^[1-9]\d{0,2}(\,\d{3})+(\.\d*)?$/
return THOUSANDS_SEPARATORS_REGEXP.test(input.trim())
```

src/main/common/utils/numericUtils.ts
```diff
@@ -10,7 +10,19 @@ export function toNumberOrUndefined (value: any): number {
...
   const strVal: string = value && value.toString().trim()
-  const numberVal: number = strVal.length > 0 ? _.toNumber(strVal) : undefined
 
-  return isNaN(numberVal) ? undefined : numberVal
+  // checks length, followed by a filter by not containing ',' or valid thousand separators
+  const numberVal: number = strVal.length > 0
+    ? (!strVal.includes(',') || containsAThousandSeparator(strVal))
+      ? numeral(strVal).value()
+      : strVal
+    : undefined
```
rlewan: Nested ternary expressions are a little hard to read, I think we could just use an `if` statement here.

src/main/app/forms/form.ts
```diff
@@ -47,19 +47,22 @@ export class Form<Model> {
...
+  rawData: object
   errors: FormValidationError[]
 
   /**
    * @param model - a object used to fill the form
+   * @param rawData - a raw data used to create model instance
    * @param errors - an array of error objects
    */
-  constructor (model: Model, errors: ValidationError[] = []) {
+  constructor (model: Model, rawData: object = undefined, errors: ValidationError[] = []) {
```
rlewan: If you added the new parameter after existing ones then you wouldn't need to assign default value to `rawData`, it would be `undefined` by default if not provided. An additional example of such approach is that you're sure you don't break any existing calls. Please change this.

src/main/app/forms/form.ts
```diff
@@ -77,6 +80,23 @@ export class Form<Model> {
       .map((error: FormValidationError) => error.message)[0]
   }
 
+  /**
+   * Get raw data for given field name.
+   * @param {string} fieldName
+   * @returns {object}
+   */
+  rawDataFor (fieldName: string): object {
```
rlewan: The logic inside this method seems to be more complex than just returning a field value. Can please you give some more info on what this method should be doing? Can you please add some unit tests for it?

src/main/common/utils/numericUtils.ts
```diff
@@ -1,5 +1,21 @@
+import * as numeral from 'numeral'
 import * as _ from 'lodash'
 
+function convert (strVal: string) {
```
rlewan: Can you please specify the `number` return value type?

src/main/common/utils/numericUtils.ts
```diff
@@ -1,5 +1,21 @@
...
+  // checks length, followed by a filter by not containing ',' or valid thousand separators
```
rlewan: I don't think this comment is needed, it's clear enough to read the method body. Instead it would be worth to explain why we use both `numeral` and `_` to convert a string to number.

src/main/common/utils/numericUtils.ts
```diff
@@ -10,7 +26,15 @@ export function toNumberOrUndefined (value: any): number {
   }
 
   const strVal: string = value && value.toString().trim()
-  const numberVal: number = strVal.length > 0 ? _.toNumber(strVal) : undefined
 
-  return isNaN(numberVal) ? undefined : numberVal
+  // convert to number
+  const numberVal = convert(strVal)
```
rlewan: You wouldn't need this comment if `convert` had a more accurate name and/or all types specified.

src/main/app/forms/form.ts
geek-soft: The method does exactly the same thing as `valueFor` just taking data from `rawData` property. Could be refactored a bit and must be tested 👍

src/main/common/utils/numericUtils.ts
```diff
@@ -10,7 +26,15 @@ export function toNumberOrUndefined (value: any): number {
...
+
+  // checks whether is null or not a number then undefined else  number
```
rlewan: This comment is not needed.

src/main/common/utils/numericUtils.ts
```diff
@@ -1,5 +1,21 @@
...
 
+function convert (strVal: string) {
+  // checks length, followed by a filter by not containing ',' or valid thousand separators
+
+  if (strVal.length > 0) {
+
+    if ((!strVal.includes(',') || containsAThousandSeparator(strVal))) {
+      return numeral(strVal).value()
+    } else {
+      return _.toNumber(strVal)
```
geek-soft: I would suggest using just one tool to convert string to number. Having two methods might be read as something we missed that when corrected will have unexpected side-effect.

If numeral library is giving us surprising result we might return undefined early when unsupported case is reached.

src/main/common/utils/numericUtils.ts
```diff
@@ -10,7 +26,15 @@ export function toNumberOrUndefined (value: any): number {
...
 
-  return isNaN(numberVal) ? undefined : numberVal
+  // convert to number
+  const numberVal = convert(strVal)
+
+  // checks whether is null or not a number then undefined else  number
+  return !numberVal || isNaN(numberVal) ? undefined : numberVal
+}
+
+function containsAThousandSeparator (input: string): boolean {
```
geek-soft: Move that method above the `convert` method.

src/test/common/utils/numericUtils.ts
```diff
@@ -32,6 +32,22 @@ describe('toNumberOrUndefined', () => {
     it('string given', () => {
       expect(toNumberOrUndefined('lalala')).to.be.eq(undefined)
     })
+
+    it('given 1110,100', () => {
```
geek-soft: A `given string 1110,100` would be more consistent. You could also remove string from other places as string should be the default in use. The cases you tests passing numeric value might have 'numeric' added to it to highlight unusual use case.

src/test/features/claim/form/models/claimAmountRow.ts
```diff
@@ -23,9 +23,23 @@ describe('ClaimAmountRow', () => {
     it('should deserialize all fields', () => {
       expect(ClaimAmountRow.fromObject({
         reason: 'Something',
-        amount: 100.01
+        amount: '100.01'
       })).to.deep.equal(new ClaimAmountRow('Something', 100.01))
     })
+
+    it('should deserialize amount containing comma', () => {
```
geek-soft: I suggest removing these 2 tests from here. The problem of deserialising numbers does not belong to that component. I would safely assume that as long as correct deserialisation is implemented we are safe, otherwise we would need to have such tests for every place where number is asked for.

src/test/features/claim/form/models/claimAmountRow.ts
```diff
@@ -82,6 +96,20 @@ describe('ClaimAmountRow', () => {
         expect(errors.length).to.equal(1)
         expectValidationError(errors, ValidationErrors.REASON_TOO_LONG)
       })
+
+      it('row with amount having invalid comma', () => {
```
geek-soft: That in turn makes sense to me to have 👍

src/test/app/forms/form.ts
```diff
@@ -105,6 +105,19 @@ describe('Form', () => {
       expect(form.valueFor('address[city]')).to.equal('London')
     })
 
+    it('should return raw value if value associated with given field fails validation', () => {
```
geek-soft: The `form.rawDataFor` function has nothing to do with validation errors. Form might have no validation error and outcome will be the same. Please correct the test body and description to remove the confusion.

src/test/common/utils/numericUtils.ts
```diff
@@ -48,28 +60,44 @@ describe('toNumberOrUndefined', () => {
       expect(toNumberOrUndefined(-10)).to.be.eq(-10)
     })
 
-    it('negative numeric string given', () => {
+    it('negative numeric given', () => {
```
geek-soft: No numeric value is passed to the `toNumberOrUndefined` function so tests should not have numeric world in description. I suggested to use numeric only when numbers are passed in (not strings)

PR #372, opened 2018-02-13
ROC-3054 Check country

robertparkinson:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-3054

### Change description
Do not allow Isle of Man postcodes

### Work checklist

- [ ] Yes
- [x] No


src/main/app/forms/validation/validators/isCountrySupported.ts
```diff
@@ -19,6 +19,8 @@ export class CheckCountryConstraint implements ValidatorConstraintInterface {
   async validate (value: any | string, args?: ValidationArguments): Promise<boolean> {
     if (value === undefined || value === null || value === '') {
       return true
+    } else if (value.toUpperCase().startsWith('IM')) {
```
timja-kainos: does the postcode lookup service have these?

robertparkinson: it returns with an error when searching for the couple I've tried

timja-kainos: then why add this?

robertparkinson: This constraint allows postcodes that are not found in the postcode lookup as it is not 100% reliable. We do know that all post codes that begin with IM are not allowed so returning false will show them an appropriate error message

geek-soft: The implementation is not very readable. Can you extract 'IM' into collection that states what post codes are blacklisted. Variable name will hopefully give us more context.

Also use regular `if` instead of `else if` since previous statement returns early.

src/test/app/forms/validation/validators/isCountrySupported.ts
```diff
@@ -71,6 +71,17 @@ describe('IsCountrySupported', () => {
 
         expect(await constraint.validate('EH9 1SH', validationArgs(Country.defendantCountries()))).to.equal(false)
       })
+
+      it('given an Isle of Man postcode', async () => {
+        nock(mockPostcode)
```
geek-soft: That should not be needed since request to the postcode lookup API is not being made in this scenario.

src/main/app/forms/validation/validators/isCountrySupported.ts
```diff
@@ -13,13 +13,22 @@ import { request } from 'client/request'
 
 const postcodeClient = new PostcodeInfoClient(config.get<string>('postcodeLookup.apiKey'), request)
 
+export enum BlockedPostcodes {
```
geek-soft: Do not export it, it is internal implementation until we have a reason to make it public.

src/main/app/forms/validation/validators/isCountrySupported.ts
```diff
@@ -13,13 +13,22 @@ import { request } from 'client/request'
...
+
 @ValidatorConstraint()
 export class CheckCountryConstraint implements ValidatorConstraintInterface {
 
   async validate (value: any | string, args?: ValidationArguments): Promise<boolean> {
     if (value === undefined || value === null || value === '') {
       return true
     }
+
+    if (value.toUpperCase().startsWith(BlockedPostcodes.ISLE_OF_MAN)) {
```
geek-soft: Shall we trim the value before doing `startsWith`? I guess there is a chance to provide postcode that starts with ` IM` (whitespace prefixed).

PR #376, opened 2018-02-14
ROC-3200: Calculate interest from date of issue

robertparkinson:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-3200

### Change description

This PR change the way interest is calculated when no specific date was given by claimant. Mark said that in such cases interest should be calculated from the date of issue (might be un the future) not from the date of case submission.

The following changes were made:
- unused and duplicated calculation logic used in response receipt has been removed
- all interest calculation functions (for both claim and draft) has been moved to one place / file
- calculate interest method has been updated to make API call only when interest period is greater or equal 1 day (that required us to remove unused interest calculation mocks from route tests)
- interest end date has been implemented to be either today or interest from date (whoever is later)
- interest end date has been added to calculation output so that can be displayed on the screen (previously today was used in all the cases) 

### Work checklist

- [ ] Yes
- [x] No


src/main/common/interest.ts
```diff
@@ -19,7 +19,8 @@ export async function getInterestDetails (claim: Claim): Promise<object> {
   }
 
   const todayDate: Moment = MomentFactory.currentDate()
-  const noOfDays: number = todayDate.diff(interestDate, 'days') + 1
+  const noOfDays: number = interestDate.diff(interestDate, 'days') === 0 ?
```
geek-soft: That makes no sense to me. Doing diff on the same date instance (`interestDate.diff(interestDate, 'days')`) will always give 0 e.g. difference between today and today is always 0.

Another thing is that if diff always evaluates to 0 then `todayDate.diff(interestDate, 'days')` is always executed which basically reverts the calculation logic to the initial version which was clarifies as not woking properly.

src/test/common/utils/interest.ts
```diff
@@ -38,23 +38,23 @@ describe('getInterestDetails', () => {
     expect(await getInterestDetails(claim)).to.be.eq(undefined)
   })
 
-  it('should return 1 for number of days when interest date starts from today', async () => {
+  it('should return 0 for number of days when interest date starts from today', async () => {
     claimStoreServiceMock.mockCalculateInterestRate(5)
 
     const today = MomentFactory.currentDate()
     const claim: Claim = new Claim().deserialize({ ...sampleClaimObj, createdAt: today })
```
geek-soft: Shouldn't we set `issuedOn` for the test case?

robertparkinson: oops! I changed the other one but forgot this one. Will change it. thanks!

src/main/common/interest.ts
```diff
@@ -15,11 +15,11 @@ export async function getInterestDetails (claim: Claim): Promise<object> {
   if (claim.claimData.interestDate.type === InterestDateType.CUSTOM) {
     interestDate = claim.claimData.interestDate.date
   } else {
-    interestDate = claim.createdAt
+    interestDate = claim.issuedOn
```
geek-soft: Are you aware that issue date can be the next day if claim issued after 4pm?

Change will cause that on the UI we would say `Interest from 14th Feb to 13th Feb`.

geek-soft: Maybe the solution is just to strip time out of creation date. This way we would align date calculation with interest calculation as calculation in claim service does not take time in consideration when interest is calculated.

robertparkinson: That is why I have switched to using the issue date as it will take into account that time difference

geek-soft: Calculating interest from issue date is much different then calculating interest from case submission date. Do you have business agreement for changing the calculation logic? Issue date might be few days after the submission date as issue date must be working day.

src/main/common/interest.ts
```diff
@@ -15,17 +16,13 @@ export async function getInterestDetails (claim: Claim): Promise<object> {
   if (claim.claimData.interestDate.type === InterestDateType.CUSTOM) {
     interestDate = claim.claimData.interestDate.date
   } else {
-    interestDate = claim.createdAt
+    interestDate = claim.issuedOn.startOf('day')
```
geek-soft: Is that needed? I think `issuedOn` should only have date information, no time there.

src/main/common/interest.ts
```diff
@@ -15,17 +16,13 @@ export async function getInterestDetails (claim: Claim): Promise<object> {
...
   }
 
-  const todayDate: Moment = MomentFactory.currentDate()
-  const noOfDays: number = todayDate.diff(interestDate, 'days') + 1
+  const todayDate: Moment = MomentFactory.currentDate().startOf('day')
```
geek-soft: Definitely not needed, `#currentDate` factory already trims time by doing

```return moment().hours(0).minutes(0).seconds(0).milliseconds(0)```

File: src/main/app/common/InterestData.ts [hunk 6]
```diff
@@ -0,0 +1,15 @@
+import { Moment } from 'moment'
+
+export class InterestData {
```
geek-soft: What about making this an interface? It does not really have to be a class at this point.

src/main/features/response/routes/claim-details.ts
```diff
@@ -4,18 +4,23 @@ import { Claim } from 'claims/models/claim'
...
 
 /* tslint:disable:no-default-export */
 export default express.Router()
   .get(Paths.claimDetailsPage.uri, async (req: express.Request, res: express.Response, next: express.NextFunction) => {
     try {
       const claim: Claim = res.locals.claim
       const feeAmount: number = MoneyConverter.convertPenniesToPounds(claim.claimData.feeAmountInPennies)
-      const interestData = await getInterestDetails(claim)
+      const interestData: InterestData = await getInterestDetails(claim)
+      const interestToDate: Moment = MomentFactory.currentDate().startOf('day').diff(interestData.interestFromDate, 'days') < 0 ?
```
geek-soft: The same here `.startOf('day')` is not needed.

geek-soft: What about moving this to `getInterestDetails` method so logic of calculating from and to date is kept close especially that the logic is more complex now.

src/main/common/interest.ts
robertparkinson: could update currentDate to use start of? it's cleaner than the current implementation?

src/main/app/common/InterestData.ts
robertparkinson: done

PR #385, opened 2018-02-21
ROC-3197 Invalidate idam token on logout

dharmendrak:


src/main/app/idam/idamClient.ts
```diff
@@ -72,4 +72,9 @@ export class IdamClient {
         )
       })
   }
+
+  static invalidateCode (code: string): Promise<void> {
```
geek-soft: Code is a bit misleading here especially that above method `exchangeCode` use the same vocabulary while the code in that case means totally different thing.

Please rename to `invalidateSession` or `invalidateUserAuthToken`.

src/test/http-mocks/idam.ts
```diff
@@ -23,6 +23,12 @@ export function resolveExchangeCode (token: string) {
     })
 }
 
+export function resolveInvalidateCode (token: string) {
```
geek-soft: When main method is renamed please do follow-up rename here as well.

src/test/routes/logout.ts
```diff
@@ -21,6 +21,7 @@ describe('Logout receiver', () => {
   describe('on GET', () => {
     it('should remove session cookie', async () => {
       idamServiceMock.resolveRetrieveUserFor('1', 'citizen')
+      idamServiceMock.resolveInvalidateCode('ABC')
```
geek-soft: Add missing test case for the following scenarios:
- session invalidation failed (500 error from the API - cookie still deleted)
- session does not exist (no session cookie - no API call, no cookie deleted)

src/main/routes/logout.ts
```diff
@@ -2,13 +2,30 @@ import * as express from 'express'
...
+    if (jwt) {
+      try {
+        await IdamClient.invalidateCode(jwt)
+      } catch (error) {
+        const { id } = JwtUtils.decodePayload(this.bearerToken)
+        logger.error(`Failed invalidating JWT for userId  ${id}`)
+      }
+    }
+
     const cookies = new Cookies(req, res)
```
geek-soft: Please move it under `if (jwt) {` condition as well. It means only when cookie exists the logout logic is executed. If session has been already destroyed (cookie removed) then only redirect happens.

src/main/routes/logout.ts
```diff
@@ -2,13 +2,30 @@ import * as express from 'express'
...
 export default express.Router()
-  .get(Paths.logoutReceiver.uri, (req: express.Request, res: express.Response, next: express.NextFunction) => {
+  .get(Paths.logoutReceiver.uri, async (req: express.Request, res: express.Response, next: express.NextFunction): Promise<void> => {
+    const jwt: string = JwtExtractor.extract(req)
+
+    if (jwt) {
+      try {
+        await IdamClient.invalidateCode(jwt)
+      } catch (error) {
+        const { id } = JwtUtils.decodePayload(this.bearerToken)
```
geek-soft: Method can throw the following error, please catch it by wrapping route handled with `ErrorHandling.apply` function. Of course that will mean that no cookie will get deleted when.decoding failed but I am OK with that since this is very rare case and probably session has been manually set to something that is not valid session / JWT anyway.

 ```
throw new Error(`Unable to parse JWT token: ${jwt}`)
```

Comment (dharmendrak) [reply src/main/routes/logout.ts]: done

File: src/main/routes/logout.ts (hunk 4 above)
Comment (dharmendrak) [reply src/main/routes/logout.ts]: done

File: src/test/routes/logout.ts (hunk 3 above)
Comment (dharmendrak) [reply src/test/routes/logout.ts]: done

File: src/test/http-mocks/idam.ts (hunk 2 above)
Comment (dharmendrak) [reply src/test/http-mocks/idam.ts]: done

File: src/main/app/idam/idamClient.ts (hunk 1 above)
Comment (dharmendrak) [reply src/main/app/idam/idamClient.ts]: done

File: src/main/app/idam/idamClient.ts [hunk 6]
```diff
@@ -72,4 +72,9 @@ export class IdamClient {
         )
       })
   }
+
+  static invalidateSession (code: string): Promise<void> {
```
geek-soft: Can you please add precondition to throw an error when jwt is blank? Something like:

```
   if (!jwt) {
      return Promise.reject(new Error('JWT is required'))
    }
```

At the same time please rename `code` to be `jwt` to be more clear.

src/main/routes/logout.ts
```diff
@@ -2,14 +2,35 @@ import * as express from 'express'
...
 export default express.Router()
-  .get(Paths.logoutReceiver.uri, (req: express.Request, res: express.Response, next: express.NextFunction) => {
-    const cookies = new Cookies(req, res)
-    cookies.set(sessionCookie, '', { sameSite: 'lax' })
-    res.redirect(Paths.homePage.uri)
-  })
+  .get(Paths.logoutReceiver.uri,
+    ErrorHandling.apply(async (req: express.Request, res: express.Response, next: express.NextFunction): Promise<void> => {
+      const jwt: string = JwtExtractor.extract(req)
+      const cookies = new Cookies(req, res)
```
geek-soft: Please move it down, right before `cookies.set`. No need to declare it sooner and it is good to keep declaration closed to where it is needed.

src/test/routes/logout.ts
```diff
@@ -21,11 +21,31 @@ describe('Logout receiver', () => {
...
+      idamServiceMock.resolveInvalidateSession('ABC')
 
       await request(app)
         .get(AppPaths.logoutReceiver.uri)
         .set('Cookie', `${cookieName}=ABC`)
         .expect(res => expect(res).to.have.cookie(cookieName, ''))
     })
+
+    it('should remove session cookie even when session invalidation is failed ', async () => {
+      idamServiceMock.resolveRetrieveUserFor('1', 'citizen')
```
geek-soft: That is something that worries me. It is not being used at all for logout endpoint so remove it.

Please also add `attachDefaultHooks(app)` as in other route tests, having this would detect unused mock.

src/test/routes/logout.ts
```diff
@@ -21,11 +21,31 @@ describe('Logout receiver', () => {
...
+      idamServiceMock.rejectInvalidateSession(idamServiceMock.defaultServiceAuthToken)
+
+      await request(app)
+        .get(AppPaths.logoutReceiver.uri)
+        .set('Cookie', `${cookieName}=${idamServiceMock.defaultServiceAuthToken}`)
```
geek-soft: It might be misleading to read that service auth token (idamServiceMock.defaultServiceAuthToken) is used in logout. I would be happy for you to rename `idamServiceMock.defaultServiceAuthToken` to `idamServiceMock.defaultAuthToken`

src/test/routes/logout.ts
dharmendrak: done

src/test/routes/logout.ts
dharmendrak: done

PR #392, opened 2018-02-26
ROC-3224: Configure App Insights

rlewan:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-3224

- [ ] Yes
- [X] No

### DIFF
--- a/src/main/server.ts
+++ b/src/main/server.ts
@@ -1,6 +1,11 @@
 #!/usr/bin/env node
 
 import './ts-paths-bootstrap'
+
+import { AppInsights } from 'modules/app-insights'
+// App Insights needs to be enabled as early as possible as it monitors other libraries as well
+AppInsights.enable()
+
 import * as config from 'config'
 import * as toBoolean from 'to-boolean'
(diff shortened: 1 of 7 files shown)


src/main/server.ts
```diff
@@ -1,6 +1,11 @@
 #!/usr/bin/env node
 
 import './ts-paths-bootstrap'
+
+import { AppInsights } from 'modules/app-insights'
+// App Insights needs to be enabled as early as possible as it monitors other libraries as well
+AppInsights.enable()
```
timja-kainos: what happens if you don't have the key set?

rlewan: Nothing, it just starts up normally.

rlewan: Or sec, there is a default for the key.

rlewan: It will fail if I remove the default.

rlewan: I guess I'll remove the default and set the dummy value for `development`, `dockertests` and `mocha`?

PR #398, opened 2018-02-27
ROC-3255 Retry requests

timja-kainos:


src/main/app/claims/claimStoreClient.ts
```diff
@@ -89,20 +91,21 @@ export class ClaimStoreClient {
...
         }
-        if (claim) {
-          return new Claim().deserialize(claim)
-        } else {
-          throw new Error('Call was successful, but received an empty claim instance')
-        }
+        return new Claim().deserialize(claim.body)
+      })
+      .catch(err => {
+        console.log(err)
```
robertparkinson: is this a WIP? if not this needs removing

timja-kainos: lint will block that dw ;), its wip but pretty much done, just been discussing with damian and rafal

package.json
```diff
@@ -77,6 +77,7 @@
     "reflect-metadata": "^0.1.10",
     "request": "^2.81.0",
     "request-promise-native": "^1.0.3",
+    "requestretry": "^1.13.0",
```
geek-soft: We need to use forked version here.

src/main/app/claims/claimStoreClient.ts
```diff
@@ -1,4 +1,4 @@
-import { request } from 'client/request'
+import { retryingRequest } from 'client/request'
```
geek-soft: Didn't we agree to use retrying request by default in all places?

config/default.json
```diff
@@ -71,6 +71,9 @@
   "http": {
     "timeout": 10000
   },
+  "requestRetry": {
+    "maxAttempts": 5
```
geek-soft: I would say 3 would be enough. Retrying 5 times on most HTTP errors (status code >= 300) seems pointless.

src/main/app/client/request.ts
```diff
@@ -3,27 +3,34 @@ import { RequestTracingHandler } from 'logging/requestTracingHandler'
...
-
+const defaultOptions = {
+  json: true,
+  timeout: timeout,
+  fullResponse: false,
+  maxAttempts: maxAttempts
+}
 const request: RequestPromiseAPI = RequestTracingHandler.proxy(
-  RequestLoggingHandler.proxy(rpWithDefaultsSet)
+  RequestLoggingHandler.proxy(requestPromise.defaults(defaultOptions))
```
geek-soft: Passing `maxAttempts: maxAttempts` to regular request object does not make much sense and might be misleading.

src/main/app/claims/claimStoreClient.ts
rlewan: As agreed it will be first applied to claim store calls only and then to the rest.

geek-soft: I am happy with using `retryingRequest` for now to emphasise different behaviour in claim-store client. Later when we apply retrying request to all other places we should rename it back to be `request`

PR #408, opened 2018-03-01
ROC-3273 Defendant sent to wrong claim

robertparkinson:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-3273

### Change description
We now redirect users to the dashboard after they have logged in. 

### Work checklist

- [ ] Yes
- [x] No


src/main/routes/receiver.ts
```diff
@@ -90,6 +91,12 @@ async function retrieveRedirectForLandingPage (user: User): Promise<string> {
   const claimIssuedButNoResponse: boolean = (claimAgainstDefendant).length > 0
     && !atLeastOneResponse
 
+  const draftResponse: Draft<ResponseDraft>[] = await new DraftService().find('response', '100', user.bearerToken, (value: any): ResponseDraft => {
```
timja-kainos: I don't think this is the fix, the claimIssuedButNoResponse case covers this.

This happens before there is a draft response

geek-soft: Behaviour before change is as follows:

```
if (claimIssuedButNoResponse) {
    return ResponsePaths.taskListPage
      .evaluateUri({ externalId: claimAgainstDefendant.pop().externalId })
  }
```

There is redirect to task list in case response has not been submitted yet but another linking happened. Robert agreed solution with Mark that in that case redirect to dashboard should happen so that user can make decision what to do, whether continue response of the draft or start response for newly linked claim.

timja-kainos: right

src/test/routes/receiver.ts
```diff
@@ -60,6 +60,7 @@ describe('Login receiver', async () => {
           claimStoreServiceMock.resolveRetrieveByClaimantIdToEmptyList()
           claimStoreServiceMock.resolveRetrieveByDefendantIdToEmptyList()
           draftStoreServiceMock.resolveFindNoDraftFound()
+          draftStoreServiceMock.resolveFindNoDraftFound()
```
geek-soft: When we use `draftStoreServiceMock.resolveFindNoDraftFound` we should pass what draft type has not been found otherwise it is a bit magic.

src/test/routes/receiver.ts
```diff
@@ -74,6 +75,7 @@ describe('Login receiver', async () => {
           draftStoreServiceMock.resolveFind('claim')
           claimStoreServiceMock.resolveRetrieveByClaimantIdToEmptyList()
           claimStoreServiceMock.resolveRetrieveByDefendantIdToEmptyList()
+          draftStoreServiceMock.resolveFindNoDraftFound()
```
geek-soft: Here I can only guess the second one is to mock request for response draft but without looking at receiver code I would not be able to tell it. Please be more specific in what response is being mocked.

src/test/routes/receiver.ts
```diff
@@ -97,6 +100,22 @@ describe('Login receiver', async () => {
         })
       })
 
+      context('when defendant has one or more draft responses', async () => {
+
+        it('should redirect to response task-list', async () => {
```
geek-soft: Test case lies here, it says redirect to task list while in fact we expect redirect to dashboard.

src/main/routes/receiver.ts
geek-soft: Please rename to `draftResponses` since list is returned.

geek-soft: I think that solution has a chance to work but we agreed with Mark to take users always to dashboard to simplify it.

src/main/routes/receiver.ts
```diff
@@ -154,14 +112,13 @@ export default express.Router()
...
-            res.redirect(await retrieveRedirectForLandingPage(user))
+            res.redirect(DashboardPaths.dashboardPage.uri)
           } else {
             Promise.all(user.getLetterHolderIdList().map(
             (letterHolderId) => linkDefendantWithClaimByLetterHolderId(letterHolderId, user)
             )
           )
-            .then(async () => res.redirect(await retrieveRedirectForLandingPage(user)))
-            .catch(async () => res.redirect(await retrieveRedirectForLandingPage(user)))
+            .then(async () => res.redirect(DashboardPaths.dashboardPage.uri))
```
geek-soft: please remove `async` keyword there, redirect does not require it.

geek-soft:
@robertparkinson please make sure we state clearly that that change just redirect everything to dashboard.

pragnesh-patel:
CodeCov is reporting: 
Coverage   61.97%   62.09%   +0.12%

PR #410, opened 2018-03-01
ROC-3255: Retry remaining requests

rlewan:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-3255

### Change description

Use `retryrequest` for all remaining clients.

- [ ] Yes
- [X] No

### DIFF
--- a/src/main/app/client/request.ts
+++ b/src/main/app/client/request.ts
@@ -22,9 +22,6 @@ const defaultRequestRetryOptions = {
   maxAttempts: maxAttempts
 }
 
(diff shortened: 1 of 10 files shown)


src/main/app/client/request.ts
```diff
@@ -37,7 +34,6 @@ const retryingRequest: RequestPromiseAPI = RequestTracingHandler.proxy(
 )
 
 export {
-  request,
-  requestNonPromise,
-  retryingRequest
+  retryingRequest as request,
```
geek-soft: That is great :)

PR #423, opened 2018-03-08
ROC-3268 Update tests for enabling CCD

timja-kainos:


src/integration-test/helpers/clients/idamClient.ts
```diff
@@ -31,13 +31,64 @@ export class IdamClient {
...
+    return request.get({
+      uri: `${baseURL}/testing-support/accounts/pin/${letterHolderId}`,
+      resolveWithFullResponse: true,
+      rejectUnauthorized: false,
+      json: false
+    })
+  }
+
+  /**
+   * Authorizes user
```
rlewan: That comment probably wasn't updated.

src/integration-test/helpers/clients/idamClient.ts
```diff
@@ -31,13 +31,64 @@ export class IdamClient {
...
+   *
+   * @param {string} email
+   * @param upliftToken the pin user's authorization header
+   * @returns {Promise<string>}
+   */
+  static async upliftUser (email: string, upliftToken: string): Promise<string> {
```
rlewan: Would you mind renaming `upliftToken` to `userAuthenticationToken` or just `authenticationToken` like it's done in other places? This is just a normal user auth token and the parameter name makes it sound like something different.

timja-kainos: its not really though, its a specific token for a pin user, and the token in this context is only used for uplifting

timja-kainos:
image

src/integration-test/helpers/clients/idamClient.ts
pragnesh-patel: pinAuthenticationToken / pinUpliftToken ?

PR #430, opened 2018-03-09
ROC-3215: Separate eligibility pages from task list

jacmurphy:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-3215

### Change description

Changes:
- Move eligibility out of claim folder and make it public (available for everyone)
- Reimplement eligibility routes using abstract class
- Save responses in a cookie
- Use cookie to make draft as eligible
- Require draft to be eligible using middleware

Diagram that documents new flow: https://tools.hmcts.net/confluence/display/~damian.dunajski/CMC+Eligibility

Note: Code is written in backward compatible way so that application can understand eligibility objects which are part of existing drafts.

- [] Yes
- [x] No


src/main/features/claim/guards/claimEligibilityGuard.ts
```diff
@@ -0,0 +1,47 @@
...
+import { DraftService } from 'services/draftService'
+
+import { CookieEligibilityStore } from 'eligibility/store'
+
+import { Paths as EligibilityPaths } from 'eligibility/paths'
+
+const eligibilityStore = new CookieEligibilityStore()
+
+export class ClaimEligibilityGuard {
+  /**
```
timja-kainos: :(, probably need a quick way to skip eligibility then, because its slow and painful to fill out

tslint.json
```diff
@@ -1,6 +1,24 @@
 {
   "extends": "tslint-config-standard",
   "rules": {
+    "member-ordering": [
+      true,
+      {
+        "order": [
```
timja-kainos: why are we overriding the default here?

geek-soft: This is let us have private methods mixed with other methods so that we can have private methods straight underneath the method that uses it.

I did not find other way to remove rule describing order of private methods other then overriding whole configuration node.

src/main/features/claim/guards/claimEligibilityGuard.ts
geek-soft: We could disable it locally by feature flag or relay on quicker mechanism of populating draft ready to use.

I also thought that we could need to shut down the service for new users by redirecting everyone to MCOL in case we had serious performance issues with overwhelming pubic beta volumes. Should not be difficult to do it in form of emergency release.

timja-kainos: emergency release is an old term,
you mean make a code change?

We need to do something about it next sprint

tslint.json
timja-kainos: sounds like too many methods if that's a problem?
I would prefer private methods separate.
I recommend reverting this, but open to hearing others opinions

src/main/features/claim/guards/claimEligibilityGuard.ts
```diff
@@ -0,0 +1,47 @@
...
+   * Makes sure that eligibility check has passed prior accessing protected pages. The eligibility is assessed by
+   * checking whether draft has been marked as eligible or whether eligible cookie exists. If none of the conditions
+   * is met then user is redirected to eligibility index page, otherwise request is let through as is.
+   *
+   * @returns {express.RequestHandler} - request handler middleware
+   */
+  static check (): express.RequestHandler {
```
rlewan: We tend to name this method `requestHandler` because it returns an Express request handler.

src/main/features/eligibility/model/claimType.ts
```diff
@@ -2,7 +2,8 @@ export class ClaimType {
 
   static readonly PERSONAL_CLAIM = new ClaimType('PERSONAL_CLAIM', 'Just myself or my organisation')
   static readonly MULTIPLE_CLAIM = new ClaimType('MULTIPLE_CLAIM', 'More than one person or organisation')
-  static readonly REPRESENTATIVE_CLAIM = new ClaimType('REPRESENTATIVE_CLAIM', 'A client - I’m their legal representative')
+  // We are not using " I’m " as the apostrophe causes errors with the cookie
```
rlewan: Maybe there is some escaping missing? Have you tried the html entity (`&apos;`) as well?

src/main/features/eligibility/model/defendantAgeOption.ts
```diff
@@ -1,4 +1,4 @@
-import { YesNoOption } from 'models/yesNoOption'
+import { YesNoOption } from 'app/models/yesNoOption'
```
rlewan: Please don't import through `app`.

rlewan: Here and in all other places.

src/main/features/eligibility/store.ts
```diff
@@ -0,0 +1,31 @@
+import { Eligibility } from 'eligibility/model/eligibility'
+import * as Cookies from 'cookies'
+import * as express from 'express'
+
+const minuteInMilliseconds = 60 * 1000
+const cookieTimeToLiveInMinutes = 10 * minuteInMilliseconds
+
+export const cookieName = 'eligibility-check'
+
+export interface EligibilityStore {
```
rlewan: This interface is implemented by one class only, I think it can be removed.

src/main/features/eligibility/store.ts
```diff
@@ -0,0 +1,31 @@
+import { Eligibility } from 'eligibility/model/eligibility'
+import * as Cookies from 'cookies'
+import * as express from 'express'
+
+const minuteInMilliseconds = 60 * 1000
+const cookieTimeToLiveInMinutes = 10 * minuteInMilliseconds
```
rlewan: This should be a configuration entry.

src/test/a11y/mocks.ts
```diff
@@ -75,6 +75,9 @@ mock('response/guards/guardFactory', {
   GuardFactory: {
     create: () => {
       return justForwardRequestHandler.requestHandler
+    },
+    createAsync: () => {
+      return justForwardRequestHandler.requestHandler
```
rlewan: Shouldn't that return an `async` function?

tslint.json
rlewan: I don't mind having private methods mixed between public ones as long as they maintain a logical order -- e.g. something Damian mentioned where the private method is straight underneath the public method that uses it.

src/main/features/claim/guards/claimEligibilityGuard.ts
geek-soft: You are right, renamed it.

src/main/features/eligibility/model/claimType.ts
geek-soft: The problem was that `’` was escaped into `&apos;` and last char (`;`) was considered as cookie delimiter which in the end made json invalid.

As a solution, by convention, I exclude `displayValue` from serialisation process.

src/main/features/eligibility/model/defendantAgeOption.ts
geek-soft: Removed from eligibility feature.

src/main/features/eligibility/store.ts
geek-soft: Good idea, done.

src/main/features/eligibility/store.ts
geek-soft: Removed

src/test/a11y/mocks.ts
geek-soft: It worked fine because both methods return just request handler which either evaluates in sync or async way and framework is able to handle both. Anyway I changed it to be more explicit:

```
    create: () => {
      return (req: express.Request, res: express.Response, next: express.NextFunction): void => {
        next()
      }
    },
    createAsync: () => {
      return async (req: express.Request, res: express.Response, next: express.NextFunction): Promise<void> => {
        next()
      }
    }
```

PR #434, opened 2018-03-12
Set NODE_ENV to production

timja-kainos:
Apparently this will make our app up to 3 times faster ^.^
https://www.dynatrace.com/news/blog/the-drastic-effects-of-omitting-node_env-in-your-express-js-applications/


pragnesh-patel:
why @timja-kainos is  Coverage   63.01%   63.38%   +0.37%

PR #446, opened 2018-03-13
ROC-3315: Handle 409 status with request retries

rlewan:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-3315

### Change description

An update to `ClaimStoreClient` that will make it return a saved claim by externalId if the initial save times out and a following retry results in 409 Conflict response.

### Work checklist

- [ ] Yes
- [X] No


src/main/app/claims/claimStoreClient.ts
```diff
@@ -15,21 +16,35 @@ export const claimStoreApiUrl: string = `${claimApiBaseUrl}/claims`
...
-          Authorization: `Bearer ${user.bearerToken}`
+          Authorization: `Bearer ${claimant.bearerToken}`
+        }
+      })
+      .then(claim => {
+        return new Claim().deserialize(claim)
+      })
+      .catch((err) => {
+        if (err.statusCode === HttpStatus.CONFLICT) {
+          return this.retrieveByExternalId(draft.document.externalId, claimant)
```
geek-soft: I would love to have some logs here stating that conflict occurred and we retrieved claim from data store.

src/main/app/claims/claimStoreClient.ts
```diff
@@ -15,21 +16,35 @@ export const claimStoreApiUrl: string = `${claimApiBaseUrl}/claims`
...
+        } else {
+          throw err
```
pragnesh-patel: As @geek-soft suggested logging here would be beneficial for support

timja-kainos: logging will already happen here @pragnesh-patel , damian suggested 2 lines up

pragnesh-patel: ok...

PR #447, opened 2018-03-14
Upgrade gulp

timja-kainos:
Locally run `npm rm --global gulp` and `npm install --global gulp-cli` to update your global gulp

Gulp was restarting the app on every integration test change, and on some files that didn't need restarts,
I upgraded it to get the series ability to make sure assets are processed before the app starts.
I've been testing it the last 30mins and seems fine


gulpfile.js
```diff
@@ -92,19 +94,21 @@ function copyA11ySniffer () {
...
   setTimeout(() => {
     livereload.listen({
       key: fs.readFileSync(path.join(__dirname, 'src', 'main', 'resources', 'localhost-ssl', 'localhost.key'), 'utf-8'),
       cert: fs.readFileSync(path.join(__dirname, 'src', 'main', 'resources', 'localhost-ssl', 'localhost.crt'), 'utf-8'),
     })
     nodemon({
       ext: 'ts js po',
-      stdout: true
+      stdout: true,
+      ignore: ['./src/integration-test/', 'src/main/public/js/lib', 'src/test']
```
geek-soft: I wonder if removing `js` from list of watched files wouldn't be better then ignoring `src/main/public/js/lib`.

timja-kainos: that would stop restarts on things like postcode-lookup.js.

PR #457, opened 2018-03-19
ROC-3310: Fix postback focus on multi-row forms

rlewan:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-3310

### Change description

Fixes an issue reported by DAC where the focus is not handled correctly after adding rows or calculating values in multi-row forms. The focus should stay on the button/link that was used to initiate the action. 

**Note:**
Added a fix in the place where DAC reported it but we have more multi-row forms now, will update the other places as well.

### Work checklist

- [ ] Yes
- [X] No


src/main/app/forms/actionForm.ts
```diff
@@ -0,0 +1,17 @@
+import { Form } from 'forms/form'
```
timja-kainos: Wouldn't a better fix to just be use javascript for adding the row?

rlewan: It's not only the row, there is the calculate total as well. Plus would need to add an XHR endpoint for each of multi-row page we have, this felt like a simpler solution.

timja-kainos: would have to see what it looks like, is it jumpy or not noticeable to the user?
It really seems like a hack we shouldn't be doing.

rlewan: Unless you meant to append html within JavaScript? Still sounds more complex I think.

src/main/features/claim/routes/amount.ts
```diff
@@ -23,10 +24,11 @@ function renderView (form: Form<ClaimAmountBreakdown>, res: express.Response): v
 
 function actionHandler (req: express.Request, res: express.Response, next: express.NextFunction): void {
   if (req.body.action) {
-    const form: Form<ClaimAmountBreakdown> = req.body
-    if (req.body.action.addRow) {
+    const form: ActionForm<ClaimAmountBreakdown> = req.body
+    if (form.action.addRow) {
       form.model.appendRow()
     }
+    form.postbackFocusTarget = ActionForm.buildPostbackFocusTarget(form.action)
```
geek-soft: Instead of resigning it we could have `get postbackFocusTarget()` method that would have logic from `ActionForm.buildPostbackFocusTarget`.

src/main/features/claim/views/amount.njk
```diff
@@ -11,6 +11,8 @@
       <form novalidate method="post">
         {{ csrfProtection(csrf) }}
 
+        <input type="hidden" id="focus-target" value="{{ form.postbackFocusTarget | default('') }}" />
```
geek-soft: I do not like the idea of having default value with empty string here. What about conditional `{% if form.action %}`? In that case we would just do `form.postbackFocusTarget`

src/main/public/js/append-row.js
```diff
@@ -0,0 +1,17 @@
+$(document).ready(function () {
+  $('input[name="action[addRow]"]').click(function () {
+    var lastRow = $('.claim-amount-row, .multiline-row').last()
+    var parent = lastRow.parent()
+    var newRow = lastRow.clone()
```
geek-soft: Will that clone errors as well? For example when amount was invalid on last row prior adding new row?

rlewan: Yes, it will 😞 Good catch.

rlewan: That makes the matter a bit more problematic.

geek-soft:
@rlewan PR says `or calculating values in multi-row forms` while this is only for adding new rows. Do we need to do something for calculating amount on amount page?

rlewan:
I'm working on the calculation right now.

rlewan:
Greep PR on dev Jenkins: https://dev-build.reform.hmcts.net/blue/organizations/jenkins/hmcts%2Fcmc-citizen-frontend/detail/PR-457/6/pipeline/

PR #459, opened 2018-03-20
ROC-3004 Terms and Conditions update

jacmurphy:
### JIRA link
https://tools.hmcts.net/jira/secure/RapidBoard.jspa?rapidView=406&projectKey=ROC&view=detail&selectedIssue=ROC-3004

### Change description
Updated:
 
Privacy Policy
Terms and Conditions
Cookies

Does this PR introduce a breaking change?

- [ ] Yes
- [x] No


src/main/views/cookies.njk
```diff
@@ -185,14 +199,14 @@
...
-          <th scope="col">{{ t('Purpose') }}</th>
+          <th scope="col">{{ t('Cookie name') }}</th>
+          <th scope="col">{{ t('What this cookie is for') }}</th>
           <th scope="col">{{ t('Expires') }}</th>
         </tr>
         </thead>
         <tbody>
         <tr>
-          <td>TSxxxxxxxx</td>
+          <td>TS01a0d005</td>
```
timja-kainos: uhh this doesn't exist

src/main/views/cookies.njk
```diff
@@ -205,6 +219,20 @@
           </td>
           <td class="text-area">{{ t('When you close your browser') }}</td>
         </tr>
+        <tr>
+          <td>X_CMC</td>
```
timja-kainos: this doesn't exist

src/main/views/macro/internalLink.njk
```diff
@@ -0,0 +1,3 @@
+{% macro internalLink(text, href) %}
```
timja-kainos: I would recommend removing this, we should be using html if it the macro doesn't give any benefit
the externalLink one gives a benefit that the opens in a new windows is always applied and applied consistently.
This just hides what's actually happen for no benefit I can see

src/main/views/privacy-policy.njk
```diff
@@ -1,122 +1,219 @@
...
+
+      <p>
+        {{ t('See our '+ internalLink('complaints procedure,', complaintsProcedureUri)
+        + ' if you want to complain about how we’ve handled your personal data.') | safe }}
+      </p>
 
       <p>
-        {{ t('While we make every effort to keep GOV.UK up to date, we don’t provide any guarantees, '
-        + 'conditions or warranties that the information will be:') }}
+        {{ t('You can also complain to the ' + internalLink('Information Commissioner’s Office' , informationCommissionerOfficeUri)
```
timja-kainos: why is this an internalLink?
normally all links off our site are externalLinks?

src/main/views/cookies.njk
jacmurphy: Its what is in the prototype we had it at one point but apparently not anymore! Ill remove it

src/main/views/cookies.njk
jacmurphy: It was what was in the prototype. What does ARRAffinity do? Ill need a description

timja-kainos: same description is fine

src/main/views/privacy-policy.njk
jacmurphy: I completely agree. I was following the prototype to the letter.

timja-kainos: following prototype to the letter is never a good idea, please confirm with Ronan/ Julian / Tarik
Although you will probably struggle to get hold of Tarik

src/main/views/macro/internalLink.njk
jacmurphy: I added it reads better and is cleaner to do:

{% set privacyPolicyUri = AppPaths.privacyPolicyPage.uri %}
  <p>
        {{ t('This includes the ' + internalLink('privacy policy', privacyPolicyUri)) | safe }}
      </p>

Then to do 

  <p> {{ t('This includes the <a href="{{ uri }}">privacy policy</a>.',
        { uri: AppPaths.privacyPolicyPage.uri }) | safe }}
      </p>

Also, if were to have a macro for externalLinks, then it would make sense and be consistent that we have one for internal links. 
Its faster to utilise and is very quick to change an internal link to external link as it holds the same structure.

timja-kainos: there's no structure to it and it would never change? 
It gives value when doing interpolation I see

src/main/views/privacy-policy.njk
jacmurphy: We need to bring this up in our retro then. If i cant trust the work given to me to copy, then it taints the whole concept of prototype.

src/main/views/cookies.njk
jacmurphy: 👍

src/main/views/cookies.njk
```diff
@@ -131,52 +131,59 @@
...
-          <th scope="col">{{ t('Purpose') }}</th>
+          <th scope="col">{{ t('Cookie name') }}</th>
+          <th scope="col">{{ t('What this cookie is for') }}</th>
           <th scope="col">{{ t('Expires') }}</th>
         </tr>
         </thead>
         <tbody>
         <tr>
-          <td>connect.sid</td>
+          <td>sessionKey</td>
```
timja-kainos: we don't have a sessionKey

timja-kainos:
Reviewed and signed off by Martin

PR #463, opened 2018-03-21
ROC-3376: Fix double escaping of output

rlewan:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-3376

### Change description

Disable [`i18n` interpolation escaping](https://www.i18next.com/interpolation.html#unescape) to avoid double escaping (nunjucks does it already).

- [ ] Yes
- [x] No


src/main/modules/i18n/index.ts
```diff
@@ -27,7 +27,7 @@ export class I18Next {
           caches: ['cookie']
         },
         interpolation: {
-          escapeValue: false
+          escapeValue: false // Escaping is already handed by Nunjucks
```
timja-kainos: handled

rlewan: @rlewan   Add a comment to explain why the change is needed  ea122d9
 @rlewan   Fix typo   ea3e9e5
😎

rlewan:
Green build on dev Jenkins: https://dev-build.reform.hmcts.net/blue/organizations/jenkins/hmcts%2Fcmc-citizen-frontend/detail/PR-463/2/pipeline/

src/main/modules/i18n/index.ts
```diff
@@ -26,6 +26,9 @@ export class I18Next {
           lookupCookie: 'lang',
           caches: ['cookie']
         },
+        interpolation: {
```
geek-soft: That could hide the issue of passing user input (with double quotes) through i18n, an issue I fixed in offers the other day.

I wonder whether it is better to leave it on globally and just disable in place where static content has to have double quotes - one place in the app so far.

rlewan: We shouldn't have double escaping in the application, Nunjucks already does it. Having double escaping and disabling it in each place where it is causing problems is unnecessary work.

PR #464, opened 2018-03-21
ROC-3359 Moment date conversion issue

dharmendrak:


src/main/app/claims/models/countyCourtJudgment.ts
```diff
@@ -10,7 +10,7 @@ export class CountyCourtJudgment {
                public paymentOption?: string,
                public paidAmount?: number,
                public repaymentPlan?: RepaymentPlan,
-               public payBySetDate?: Moment,
```
timja-kainos: this isn't right, we should be using a moment so we can do date work on it...

src/test/app/claims/models/settlement.ts
```diff
@@ -51,7 +51,7 @@ describe('Settlement', () => {
       const offer: Offer = myInput.partyStatements[0].offer
       const actual: Settlement = new Settlement().deserialize(myInput)
 
-      expect(actual.getDefendantOffer().completionDate.format('YYYY-DD-MM')).to.be.eq(offer.completionDate)
+      expect(actual.getDefendantOffer().completionDate).to.be.eq(offer.completionDate)
```
dharmendrak: it is doing String comparison here, is it good to do for date?

src/main/app/claims/models/countyCourtJudgment.ts
rlewan: 👍

src/main/app/claims/models/offer.ts
```diff
@@ -1,20 +1,16 @@
-
-import { Moment } from 'moment'
-import { MomentFactory } from 'common/momentFactory'
-
 export class Offer {
   content: string
-  completionDate: Moment
+  completionDate: string
```
rlewan: Same here.

src/main/app/claims/offerClient.ts
```diff
@@ -4,42 +4,45 @@ import { Claim } from 'claims/models/claim'
 import { User } from 'idam/user'
 import { Offer as OfferForm } from 'features/offer/form/models/offer'
 import * as config from 'config'
-import { request } from 'client/request'
+import { request as requestPromiseApi, RequestPromiseAPI } from 'client/request'
 
 export const claimStoreApiUrl: string = `${config.get<string>('claim-store.url')}/claims`
 
 export class OfferClient {
+  constructor (private request: RequestPromiseAPI = requestPromiseApi) {
```
rlewan: Why do you need to do this? I can't see any places where you initialise the client with something else.

dharmendrak:
wrong fix

PR #465, opened 2018-03-21
ROC-3359 Resolve moment date conversion issue

dharmendrak:


src/main/app/forms/models/localDate.ts
```diff
@@ -52,7 +52,7 @@ export class LocalDate {
   }
 
   toMoment (): moment.Moment {
-    return moment({ year: this.year, month: this.month - 1, day: this.day }) // Moment months are zero indexed
+    return moment.utc({ year: this.year, month: this.month - 1, day: this.day }) // Moment months are zero indexed
```
timja-kainos: We've done this before, the fix is to set hours minutes and sections to 0 not using .utc
```
 return moment().hours(0).minutes(0).seconds(0).milliseconds(0)
```
See MomentFactory.currentDate

dharmendrak: no does't work with this.

pragnesh-patel: 👍

geek-soft: @dharmendrak please confirm that it works still fine when BST timezone is in use. To simulate that please set both your frontend and backend to use BST timezone, submit date of `2018-01-01` and check that the same date is saved in DB.

PR #469, opened 2018-03-21
ROC-3381 Add change link for individual contact person

dharmendrak:


src/main/features/claim/views/check-and-send.njk
```diff
@@ -19,7 +19,8 @@
           changeLinks = {
             name: paths.claimantIndividualDetailsPage.uri,
             address: paths.claimantIndividualDetailsPage.uri,
-            dateOfBirth: paths.claimantDateOfBirthPage.uri
+            dateOfBirth: paths.claimantDateOfBirthPage.uri,
+            contactPerson: paths.claimantOrganisationDetailsPage.uri
```
timja-kainos: what happens for individual type? as it doesn't have a contact person, is that handled?

dharmendrak: yes it is handled in cmc-common-frontend ``` {% if (partyDetails.type === 'company' or partyDetails.type === 'organisation') %}
    {% if partyDetails.contactPerson %}
      {{ row('Contact person:', partyDetails.contactPerson, changeLinks.partyDetails or changeLinks.contactPerson) }}
    {% endif %}
  {% endif %}```

timja-kainos:
We aren't sure on whether this should be there, needs design and business input deferring to after public beta

PR #476, opened 2018-03-22
ROC-3400 Resolve update link for party details based on their type

dharmendrak:


pragnesh-patel:
Tested Locally
Need to test on AAT

PR #478, opened 2018-03-22
ROC-3215 Update eligibility cookie ttl to 2 days

timja-kainos:
2 days is the length of time idam activation emails are valid for, Martin requested this config value to be updated

### DIFF
--- a/src/main/views/cookies.njk
+++ b/src/main/views/cookies.njk
@@ -149,7 +149,7 @@
           <td>
             {{ t('Stores answers to eligibility questions') }}
           </td>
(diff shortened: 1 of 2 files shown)


src/main/views/cookies.njk
```diff
@@ -149,7 +149,7 @@
           <td>
             {{ t('Stores answers to eligibility questions') }}
           </td>
-          <td class="text-area">{{ t('Ten minutes') }}</td>
+          <td class="text-area">{{ t('48 hours') }}</td>
```
timja-kainos: has been reviewed by Ronan

PR #485, opened 2018-03-23
ROC-3407: Remove residual task list links after handoff pages

rattur:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-3407

### Change description

After handoff pages are established in the defence journey, when the user goes back to the dashboard and clicks on the claim reference number the  `When will you pay?` and 'Free Mediation' links are removed from the task list.

This PR also hides `Why do you disagree with this claim` in the following scenario:
 1. Defendant chooses 'I've already paid'
 2. Defendant chooses 'I've paid the total amount'
 
### Work checklist

- [ ] Yes
- [x] No


geek-soft:
@rattur I think deleting disabled tasks from the task list builder would be better solution. It is confusing what should be there and what should not. The way you solved the problem does not necessarily guarantee everything will work fine when we remove hand off pages, probably the code will need to be reworked anyway.

rattur:
Yes I think removing inactive code would be a cleaner, stable solution particularly as new scenarios are being addressed within the defence journey

src/main/features/response/tasks/oweMoneyTask.ts
```diff
@@ -11,7 +11,7 @@ export class OweMoneyTask {
       case ResponseType.FULL_ADMISSION:
```
yarpo: Why not simplify it?

```
case ResponseType.FULL_ADMISSION:
case ResponseType.PART_ADMISSION:
    return true
```

PR #488, opened 2018-03-23
ROC-3370 Update for service.gov.uk hostname

timja-kainos:
To be merged on monday


infrastructure/main.tf
```diff
@@ -41,6 +41,7 @@ module "citizen-frontend" {
   is_frontend  = true
   subscription = "${var.subscription}"
   additional_host_name = "${var.external_host_name}"
+  https_only = "true"
```
geek-soft: Would `https_only = true` work? (boolean instead of string)

timja-kainos: nope. module is inconsistent github.com/hmcts/moj-module-webapp README

geek-soft: Thanks for the link

Changing subject, I just realised we removed HPKP on CNP, something we need to enable in public BETA. Not necessarily first days.

geek-soft: Yeah lets see whether Google sticks to their plan. I think so far only they made that move.

PR #492, opened 2018-03-26
Add interest type to nunjucks environment

timja-kainos:


timja-kainos:
Vivek said merge then disappeared, time sensitive currently

PR #493, opened 2018-03-26
Resolve interest end date to be based on issued date

dharmendrak:


src/main/common/momentFactory.ts
```diff
@@ -6,7 +6,7 @@ export class MomentFactory {
   }
 
   static currentDate (): moment.Moment {
-    return moment().hours(0).minutes(0).seconds(0).milliseconds(0)
+    return moment.utc().hours(0).minutes(0).seconds(0).milliseconds(0)
```
geek-soft: That is not safe IMO. It means all dates will be sent to backend as UTC e.g. (2018-03-26T10:37:42.851+00:00), then if that is mapped to `LocalDateTime` on the backend side we would store `2018-03-26T10:37:42.851+00:00` instead of `2018-03-26T11:37:42.851+00:00`

UTC is safe for our custom local date class but not for everything.

dharmendrak: This method just handles date not with Time, are all mapped against LocalDate

src/main/app.ts
```diff
@@ -76,6 +77,10 @@ if (toBoolean(config.get<boolean>('featureToggles.testingSupport'))) {
   new TestingSupportFeature().enableFor(app)
 }
 
+moment.prototype.toISOString = function () {
```
geek-soft: @dharmendrak can you please add comment explaining why we doing this? Someone can be very surprised we overriding default implementation.

src/main/common/interestUtils.ts
```diff
@@ -13,13 +13,19 @@ import { InterestType as ClaimInterestType } from 'claims/models/interestType'
 import { YesNoOption } from 'models/yesNoOption'
 import { InterestTypeOption } from 'claim/form/models/interestType'
 
+function getInterestEndDate (issuedOn: moment.Moment, interestFromDate: moment.Moment): moment.Moment {
+  const currentDate = MomentFactory.currentDate()
+  const endDate = issuedOn > currentDate ? issuedOn : currentDate
```
rlewan: Why do we need this hack? Cannot we just use issueDate?

dharmendrak: no, this fix is for  weekend issue, 
if claims is issued after 4 PM on Friday, then date should be displayed as issue date, not current date and after weekend it will be current date

rlewan: But if it's not a weekend then current date is issue date.

src/main/common/interestUtils.ts
```diff
@@ -13,13 +13,19 @@ import { InterestType as ClaimInterestType } from 'claims/models/interestType'
...
+}
+
 export async function getInterestDetails (claim: Claim): Promise<InterestData> {
   if (claim.claimData.interest.type === ClaimInterestType.NO_INTEREST || claim.claimData.interest.type === undefined) {
     return undefined
   }
 
   const interestFromDate: moment.Moment = getInterestDateOrIssueDate(claim)
-  const interestToDate: moment.Moment = moment.max(interestFromDate, MomentFactory.currentDate())
+  const interestToDate: moment.Moment = getInterestEndDate(claim.issuedOn, interestFromDate)
```
rlewan: `getInterestEndDate` should be removed and we should just use `claim.issuedOn` here.

timja-kainos: @rlewan we don't stop calculating interest in the app though, only PDFs stop calculating at issue date.
so it should use either today, or if issue date is in the future use issue date

pragnesh-patel:
@dharmendrak please address changes requested before coming to QA approval

src/main/common/interestUtils.ts
dharmendrak: shouldn't for daily interestRate, date should be increasing everyday
by putting interesetEndDate to issuedOn we are fixing it one date

src/main/common/interestUtils.ts
rlewan: The other thread clarified it a bit for me. Still, the end date in that case should be a `max` between current date and issue date. We shouldn't be interest in from date at all.

PR #507, opened 2018-04-04
ROC-3147: Add End2End test

vianvi:
- [x] Yes
- [ ] No

### DIFF
--- a/src/integration-test/data/test-data.ts
+++ b/src/integration-test/data/test-data.ts
@@ -6,6 +6,8 @@ export const SMOKE_TEST_CITIZEN_USERNAME = process.env.SMOKE_TEST_CITIZEN_USERNA
 export const SMOKE_TEST_USER_PASSWORD = process.env.SMOKE_TEST_USER_PASSWORD
 
 export const claimFee = 25.00
+export const fixedInterestAmount = 100
+export const dailyInterestAmount = 5
 
 export const claimAmount: Amount = {
   type: 'breakdown',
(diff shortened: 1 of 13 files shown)


src/integration-test/data/test-data.ts
```diff
@@ -64,6 +69,21 @@ export function createClaimData (claimantType: PartyType, defendantType: PartyTy
       type: 'submission'
     }
   }
+
+  if (claimInterestBreakdown) {
```
geek-soft: Instead of having `claimInterest` and `claimInterestBreakdown` I would prefer to have `InterestType` enum, if set we either instantiate standard interest or interest breakdown.

Someone can assist you with this if you are not familiar with enums.

src/integration-test/tests/citizen/defence/pages/defendant-claim-details.ts
```diff
@@ -13,9 +13,14 @@ export class DefendantClaimDetails {
     I.click('View claim')
   }
 
-  checkClaimData (claimReference: string): void {
+  checkClaimData (claimReference: string, getTotal: boolean = true, getTotalWithInterest: boolean = false): void {
     I.see(claimReference)
-    I.see(claimAmount.getTotal())
+    if (getTotal) {
```
geek-soft: That solution is not very flexible. Passing bunch of true / false flags impacts readability as well. If this is check function we should pass claim object to it and expect that UI shows expected values.

PR #508, opened 2018-04-04
ROC-3458: no cache added

yarpo:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-3458

### Change description
No cache

### Work checklist

- [ ] Yes
- [x] No

### DIFF
--- a/src/main/app.ts
+++ b/src/main/app.ts
@@ -24,7 +24,6 @@ import { Feature as OfferFeature } from 'offer/index'
 import { TestingSupportFeature } from 'testing-support/index'
 import * as toBoolean from 'to-boolean'
 import { FeatureToggles } from 'utils/featureToggles'
-import * as noCache from 'nocache'
 
 export const app: express.Express = express()
 
@@ -39,7 +38,6 @@ app.locals.ENV = env
 
 const developmentMode = env === 'development'
 
-app.use(noCache())
 app.use(RequestTracing.middleware)
 
 const i18next = I18Next.enableFor(app)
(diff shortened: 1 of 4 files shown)


src/main/app.ts
```diff
@@ -38,6 +39,7 @@ app.locals.ENV = env
 
 const developmentMode = env === 'development'
 
+app.use(noCache())
```
geek-soft: Please use `app.use(helmet.noCache())` instead. Helmet also brings `nocache` dependency, no need to add it manually.

geek-soft: Also best to add this to `Helmet.enableFor` in `modules/helmet/index.ts`.

yarpo: ok

yarn.lock
```diff
@@ -5357,7 +5357,7 @@ nise@^1.2.0:
     path-to-regexp "^1.7.0"
     text-encoding "^0.6.4"
 
-nocache@2.0.0:
+nocache@^2.0.0:
```
geek-soft: Can you please revert that change?

src/main/modules/helmet/index.ts
```diff
@@ -19,6 +19,7 @@ export class Helmet {
 
   enableFor (app: express.Express) {
     app.use(helmet())
+    app.all(/^.+(?!\/public\/.*)$/, helmet.noCache())
```
geek-soft: I suggest `^\/(?!public).*$`. Your regex matches `/public/stylesheets/application.css` which means that that file would not get cached.

PR #510, opened 2018-04-04
ROC-3460: Use single instrumentation key for all services

rlewan:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-3460

### Change description

A change to start using a single App Insights insights instance which aggregates all services per environment.


Jenkinsfile_parameterized
```diff
@@ -1,6 +1,6 @@
 #!groovy
 //noinspection GroovyUnusedAssignment
-@Library("Infrastructure") _
+@Library("Infrastructure@feature/ROC-3460-One-app-insights-per-product") _
```
rlewan: Will update after the referenced part is merged.

infrastructure/main.tf
```diff
@@ -33,12 +33,13 @@ locals {
 }
 
 module "citizen-frontend" {
-  source = "git@github.com:contino/moj-module-webapp.git"
+  source = "git@github.com:contino/moj-module-webapp.git?ref=feature/ROC-3460-One-app-insights-per-product"
```
rlewan: Will update after the referenced part is merged.

PR #518, opened 2018-04-06
ROC-3511: Add undefined or null check in evaluateuri

sabahirfan:
### JIRA link
[ROC-3511](https://tools.hmcts.net/jira/browse/ROC-3511)

### Change description
This PR adds additional checks as mentioned in [#515](https://github.com/hmcts/cmc-citizen-frontend/pull/515#discussion-diff-179728282R54) 

### Work checklist

- [ ] Yes
- [x] No

### DIFF
--- a/src/main/common/router/routablePath.ts
+++ b/src/main/common/router/routablePath.ts
@@ -1,3 +1,5 @@
+import { StringUtils } from 'utils/stringUtils'
+
 const pathParameterRegex = /\/:[^\/]+/g
 
 export class RoutablePath {
@@ -33,6 +35,10 @@ export class RoutablePath {
     return `${featureName}/views/${viewPath}`
   }
 
+  static isValidParameterValue (parameterValue: string): boolean {
+    return !(StringUtils.isBlank(parameterValue) || parameterValue === 'undefined' || parameterValue === 'null')
+  }
+
   evaluateUri (substitutions: { [key: string]: string }): string {
     if (substitutions === undefined || Object.keys(substitutions).length === 0) {
       throw new Error('Path parameter substitutions are required')
@@ -40,6 +46,10 @@ export class RoutablePath {
 
     const path = Object.entries(substitutions).reduce((uri: string, substitution: [string, string]) => {
       const [parameterName, parameterValue] = substitution
+
+      if (!RoutablePath.isValidParameterValue(parameterValue)) {
+        throw new Error(`Path parameter :${parameterName} is not provided`)
+      }
 
       const updatedUri: string = uri.replace(`:${parameterName}`, parameterValue)
       if (updatedUri === uri) {
--- a/src/main/common/router/routablePath.ts
+++ b/src/main/common/router/routablePath.ts
@@ -2,7 +2,7 @@ import { StringUtils } from 'utils/stringUtils'
 
 const pathParameterRegex = /\/:[^\/]+/g
 
-export function isValidParameterValue (parameterValue: string): boolean {
+function isValidParameterValue (parameterValue: string): boolean {
   return !(StringUtils.isBlank(parameterValue) || parameterValue === 'undefined' || parameterValue === 'null')
 }
 
--- a/src/main/common/router/routablePath.ts
+++ b/src/main/common/router/routablePath.ts
@@ -2,6 +2,10 @@ import { StringUtils } from 'utils/stringUtils'
 
 const pathParameterRegex = /\/:[^\/]+/g
 
+export function isValidParameterValue (parameterValue: string): boolean {
+  return !(StringUtils.isBlank(parameterValue) || parameterValue === 'undefined' || parameterValue === 'null')
+}
+
 export class RoutablePath {
   private _uri: string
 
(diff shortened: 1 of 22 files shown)


src/main/common/router/routablePath.ts
```diff
@@ -33,6 +35,10 @@ export class RoutablePath {
     return `${featureName}/views/${viewPath}`
   }
 
+  static isValidParameterValue (parameterValue: string): boolean {
```
geek-soft: That makes this method public. Please move it outside class.

src/main/common/router/routablePath.ts
```diff
@@ -41,6 +47,10 @@ export class RoutablePath {
     const path = Object.entries(substitutions).reduce((uri: string, substitution: [string, string]) => {
       const [parameterName, parameterValue] = substitution
 
+      if (!RoutablePath.isValidParameterValue(parameterValue)) {
+        throw new Error(`Path parameter :${parameterName} is not provided`)
```
geek-soft: It might be provided but invalid. Shall we say for example 'Path parameter :foo is invalid'

src/test/common/router/routablePath.ts
```diff
@@ -65,4 +72,32 @@ describe('RoutablePath', () => {
       })
     })
   })
+
+  describe('isInValidParameterValue', () => {
```
geek-soft: When you move it to outside class we will need to remove that test. Thats is alright, you testing this in other tests.

src/main/common/router/routablePath.ts
yarpo: you can also make it `private`

sabahirfan: done

src/test/common/router/routablePath.ts
sabahirfan: done

src/main/common/router/routablePath.ts
sabahirfan: done

src/main/common/router/routablePath.ts
```diff
@@ -1,5 +1,11 @@
+import { StringUtils } from 'utils/stringUtils'
+
 const pathParameterRegex = /\/:[^\/]+/g
 
+function isValidParameterValue (parameterValue: string): boolean {
+  return !(StringUtils.isBlank(parameterValue) || parameterValue === 'undefined' || parameterValue === 'null')
```
rlewan: What's the purpose of `StringUtils.isBlank` if you are still checking for null and undefined later? Might as well check string length.

src/main/common/router/routablePath.ts
```diff
@@ -41,6 +47,10 @@ export class RoutablePath {
     const path = Object.entries(substitutions).reduce((uri: string, substitution: [string, string]) => {
       const [parameterName, parameterValue] = substitution
 
+      if (!isValidParameterValue(parameterValue)) {
+        throw new Error(`Path parameter :${parameterName} is invalid`)
```
rlewan: Maybe would be good to place `'` quotes around `${parameterName}` so that it's easier to see it is blank?

src/main/common/router/routablePath.ts
geek-soft: `StringUtils.isBlank` checks for `undefined` and `null` and then later we check for `'undefined'` and `'null'` (stings).

src/main/common/router/routablePath.ts
geek-soft: Param name should never be blank, we getting param names from names of the object literal properties.

rlewan: Why are you calling if `StringUtils.isBlank` on it if it's never blank?

src/test/common/router/routablePath.ts
```diff
@@ -34,6 +34,13 @@ describe('RoutablePath', () => {
       expect(new RoutablePath('/case/:id/payment/:payment-type').evaluateUri({ id: '999', 'payment-type': 'card' }))
         .to.be.equal('/case/999/payment/card')
     })
+
+    it('should fail when path parameter placeholder has invalid value', () => {
+      [{ id: undefined }, { id: null }, { id: 'null' }, { id: 'undefined' }, { id: '' }].forEach(invalidValue => {
```
timja-kainos: null or undefined are valid names and there are people with the last name of null let's not get to smart.
Please remove these checks

timja-kainos: hmm, since this is only in uri's i'll accept it, if a comment is added explaining it.
i.e. no one should be copy pasting that check elsewhere

sabahirfan: Added comment to make it more clear.

src/main/common/router/routablePath.ts
```diff
@@ -1,5 +1,16 @@
+import { StringUtils } from 'utils/stringUtils'
+
 const pathParameterRegex = /\/:[^\/]+/g
 
+/**
+ *  This method validates the path parameter value used in URI paths.
```
timja-kainos: Remove `This method` please

sabahirfan: done

PR #522, opened 2018-04-10
ROC-3510: Submit shows in dashboard for all digitally supported scenarios

varma02:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-3510

### Change description
Show submit for all digitally supported scenarios


src/main/features/response/draft/responseDraft.ts
```diff
@@ -125,6 +125,10 @@ export class ResponseDraft extends DraftDocument {
       && this.howMuchPaidClaimant.option === HowMuchPaidClaimantOption.AMOUNT_CLAIMED
   }
 
+  public isDigitalResponseSelected (): boolean {
```
geek-soft: @varma02 implementation is not right + it is difficult to read.

You would probably understand it yourself if you put some unit tests.

src/main/features/response/helpers/taskListBuilder.ts
```diff
@@ -80,7 +80,7 @@ export class TaskListBuilder {
 
   static buildSubmitSection (draft: ResponseDraft, externalId: string): TaskList {
     const tasks: TaskListItem[] = []
-    if (draft.isResponseRejectedFullyWithDispute() || draft.isResponseRejectedFullyWithAmountClaimedPaid()) {
+    if (draft.isDigitalResponseSelected()) {
```
geek-soft: @varma02 your rightfully found this places as the case of the issue but solution is wrong. 

The reason it does not work as expected is that `draft.isResponseRejectedFullyWithDispute() || draft.isResponseRejectedFullyWithAmountClaimedPaid()` evaluates to false when response is not selected which is initial state. We would like to show `Check and send` task when response is not selected.

src/main/features/response/helpers/taskListBuilder.ts
```diff
@@ -80,7 +80,7 @@ export class TaskListBuilder {
 
   static buildSubmitSection (draft: ResponseDraft, externalId: string): TaskList {
     const tasks: TaskListItem[] = []
-    if (draft.isResponseRejectedFullyWithDispute() || draft.isResponseRejectedFullyWithAmountClaimedPaid()) {
```
geek-soft: What if we just said?

```
if (!this.isResponsePopulated()
     || draft.isResponseRejectedFullyWithDispute()
     || draft.isResponseRejectedFullyWithAmountClaimedPaid())
```

src/test/features/response/helpers/taskListBuilder.ts
```diff
@@ -99,15 +99,18 @@ describe('Defendant response task list builder', () => {
     })
 
     describe('"Check and submit your response" task', () => {
+      let stub: sinon.SinonStub
       let isResponseRejectedFullyWithDisputePaidStub: sinon.SinonStub
       let isResponseRejectedFullyWithAmountClaimedPaidStub: sinon.SinonStub
 
       beforeEach(() => {
+        stub = sinon.stub(ResponseDraft.prototype, 'isResponsePopulated')
```
geek-soft: Please rename it to something more meaningful. To follow convention I would suggest `isResponsePopulatedStub `

PR #529, opened 2018-04-11
Deserialise YesNoOption as undefined if unknown option selected

geek-soft:
### JIRA link

None

### Change description

This PR changes the way unknown Yes/No option is deserialised. Throwing error from deserialisation method makes validation middleware fail with unhandled promise rejection.

### Work checklist

- [ ] Yes
- [x] No

### DIFF
--- a/src/main/app/models/yesNoOption.ts
+++ b/src/main/app/models/yesNoOption.ts
@@ -13,7 +13,7 @@ export class YesNoOption {
     } else if (input === 'no') {
       return YesNoOption.NO
     } else {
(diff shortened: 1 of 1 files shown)


src/main/app/models/yesNoOption.ts
```diff
@@ -13,7 +13,7 @@ export class YesNoOption {
     } else if (input === 'no') {
```
yarpo: why don't we extract "yes" and "no" to const and reuse it. Both are used twice in this file.

PR #533, opened 2018-04-13
Add lodash as a dependency explicitly

rlewan:
### Change description

We don't have `lodash` as our dependency but rely on it (we only have `@types/lodash`). Some builds have failed on CNP, seems like `lodashah` is installed differently there:

screen shot 2018-04-13 at 09 22 21

Adding it explicitly to see if this resolves the issue.


timja-kainos:
https://github.com/hmcts/cmc-citizen-frontend/pull/532

PR #552, opened 2018-04-27
ROC-2974: Remove app/* alias and update all imports

ellie-harrison:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-2974

### Change description
Removed TS alias "app/*" and updated all imports to follow a consistent style.

### Work checklist

- [ ] Yes
- [ ] No


tsconfig.json
```diff
@@ -12,91 +12,101 @@
...
       ],
       "drafts/*": [
-        "app/drafts/*"
+        "main/app/drafts/*"
       ],
       "claims/*": [
-        "app/claims/*"
+        "main/app/claims/*"
+      ],
+      "app-common/*": [
```
timja-kainos: what do you think about using either `common` or `shared` instead of `app-common` I'm not a big fan of the combined name?

ellie-harrison: `common` would have conflicts with another alias, but `shared` would certainly be a better name.

geek-soft: Happy with `shared` to start with. In the future we would like to remove app directory so the need for having both `common` and `shared` might disappear.

ellie-harrison:
I have updated the `app-common` to be `common` (as it is the most commonly used of the two directories called common) and the conflicting `common` to be `shared`.

Additionally I have added 2 new aliases for the files that were imported using 'main' to tidy those up a bit.

PR #554, opened 2018-04-27
Update govuk_template_jinja to the latest version 🚀

robertparkinson:
Closes #535

### Work checklist

- [ ] Yes
- [ ] No


package.json
```diff
@@ -63,7 +63,7 @@
     "es6-promisify": "^5.0.0",
     "express": "^4.13.3",
     "govuk-elements-sass": "^3.1.2",
-    "govuk_template_jinja": "0.23.0",
+    "govuk_template_jinja": "0.23.3",
```
timja-kainos: yarn.lock changes missing and please either check the template has been updated to match the changes in the template version, or remove the template we use and use the packaged one (not the smallest bit of work)

robertparkinson: have just pushed the lock file now. will compare the templates

robertparkinson: they just added some new css and changed some fonts by the looks of it. nothing to do on our side

geek-soft:
This needs to be well tested before merging as it gives us few CSS / JS assets. Also we should look into differences in the template itself, since we modified it.

robertparkinson:
Have already compared template as mentioned above with tims comment. Shall create a story for greater testing

src/main/views/govukTemplate.njk
```diff
@@ -31,100 +30,101 @@
...
   <meta name="viewport" content="width=device-width, initial-scale=1">
-  <meta property="og:image" content="{{ asset_paths['images_vendor'] }}/opengraph-image.png?{{ govuk_template_version }}">
 
   {% block head %}{% endblock %}
+
+  <meta property="og:image" content="{{ asset_paths['images_vendor'] }}/opengraph-image.png?{{ govuk_template_version }}">
 </head>
 
 <body class="{% block body_classes %}{% endblock %}">
+  <script>document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');</script>
```
timja-kainos: @geek-soft won't this get blocked by CSP? This is why it was removed as the js we were using wasn't using it. Can also be added if it has a hash of the js

PR #562, opened 2018-05-03
Add infra local cache

timja-kainos:
Reduce infra recycle time for hardware issues
This is to hopefully reduce some timeout issues in app insights, microsoft have recommended adding it

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -47,7 +47,7 @@ locals {
 }
 
 module "citizen-frontend" {
-  source = "git@github.com:contino/moj-module-webapp.git"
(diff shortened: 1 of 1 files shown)


infrastructure/main.tf
```diff
@@ -47,7 +47,7 @@ locals {
 }
 
 module "citizen-frontend" {
-  source = "git@github.com:contino/moj-module-webapp.git"
+  source = "git@github.com:hmcts/moj-module-webapp.git?ref=RPE-389/local-cache"
```
geek-soft: My only concern is that we using something that has not been merged into stable codebase and will be used in production. On one hand it worries it that it might not been properly tested. On other hand that somebody might change something on feature branch without realising that services are using it in prod.

PR #563, opened 2018-05-03
ROC-3690: Remove relative imports from tests

andrewwa-kainos:
Introduce test/* alias and replace *.ts files' imports to use it where possible

### JIRA link
https://tools.hmcts.net/jira/browse/ROC-3690

### Change description
tsconfig.json now defines a test/* alias
test *.ts files are updated to use the new alias where possible

- [ ] Yes
- [x] No


src/test/features/response/routes/statement-of-means/what-you-need.ts
```diff
@@ -2,7 +2,7 @@ import { expect } from 'chai'
 import * as request from 'supertest'
 import * as config from 'config'
 
-import { attachDefaultHooks } from '../../../../routes/hooks'
+import { attachDefaultHooks } from 'test/routes/hooks'
 import { checkAuthorizationGuards } from '../checks/authorization-check'
 
 import * as idamServiceMock from '../../../../http-mocks/idam'
```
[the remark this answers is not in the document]
andrewwa-kainos: Yep, for repetitive tasks like this one I normally commit a bit to show what I'm doing / how I'm doing it and then if that's accepted I commit the rest. I'll update the PR now

geek-soft:
@andrewwa-kainos , I made few extra commits to remove `./` relative imports. I also removed TS aliases that simply map to source folders.

geek-soft:
@andrewwa-kainos can I ask you to prepare the same change for legal frontend as well?

tsconfig.json
```diff
@@ -110,9 +109,6 @@
       ],
       "services/*": [
         "main/services/*"
-      ],
-      "integration-test/*": [
-        "integration-test/*"
```
rlewan: Why did this get removed?

geek-soft: Because we changed base URL to be `"baseUrl": "src"`. Having this we do not need aliases that points to the same directory because it is already guaranteed by base URL.

PR #568, opened 2018-05-08
Update content security policy to include new inline js

robertparkinson:
### Change description
New inline js added from govukTemplate changes needs to be included in our CSP checks

### Work checklist

- [ ] Yes
- [x] No

### DIFF
--- a/src/main/modules/helmet/modules/contentSecurityPolicy.ts
+++ b/src/main/modules/helmet/modules/contentSecurityPolicy.ts
@@ -9,7 +9,7 @@ export class ContentSecurityPolicy {
   constructor (public developmentMode: boolean) {}
 
   enableFor (app: express.Express) {
-    const scriptSrc = [self, '*.google-analytics.com']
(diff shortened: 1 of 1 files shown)


src/main/modules/helmet/modules/contentSecurityPolicy.ts
```diff
@@ -9,7 +9,7 @@ export class ContentSecurityPolicy {
   constructor (public developmentMode: boolean) {}
 
   enableFor (app: express.Express) {
-    const scriptSrc = [self, '*.google-analytics.com']
+    const scriptSrc = ['\'sha256-+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU=\'', '\'sha256-G29/qSW/JHHANtFhlrZVDZW1HOkCDRc78ggbqwwIJ2g=\'', self, '*.google-analytics.com']
```
rlewan: Perhaps it would be good to add a comment here which tells which inline scripts these hashes refer to?

robertparkinson: ah true. I should extract them to meaningful variables. will do so now

robertparkinson: have pushed some changes

geek-soft:
@robertparkinson can you do the same on legal-frontend side?

PR #575, opened 2018-05-10
ROC-3550: Update demo on changes to master

rlewan:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-3550

### Change description

Added an `afterCheckout` hook to the pipeline to update the demo branches with latest master changes.

- [ ] Yes
- [X] No

### DIFF
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -29,7 +29,7 @@ withPipeline("nodejs", product, component) {
           git push --force origin HEAD:demo
         """
       } catch (err) {
-        // Send a notification to Slack channel?
+        notifyBuildEvent channel: '#cmc-tech-notification', color: 'warning', message: 'Failed to update demo branch'
       }
     }
   }
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -21,6 +21,19 @@ static LinkedHashMap<String, Object> secret(String secretName, String envVar) {
 }
 
 withPipeline("nodejs", product, component) {
+  afterCheckout {
+    if (env.BRANCH_NAME == "master") {
+      try {
+        sh """
+          git fetch origin demo:demo
+          git push --force origin HEAD:demo
+        """
+      } catch (err) {
+        // Send a notification to Slack channel?
+      }
+    }
+  }
+
   after('build') {
     sh 'yarn setup'
   }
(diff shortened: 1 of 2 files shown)


Jenkinsfile_CNP
```diff
@@ -21,6 +21,19 @@ static LinkedHashMap<String, Object> secret(String secretName, String envVar) {
 }
 
 withPipeline("nodejs", product, component) {
+  afterCheckout {
+    if (env.BRANCH_NAME == "master") {
```
timja-kainos: can you use `onMaster` in here?

Jenkinsfile_CNP
```diff
@@ -21,6 +21,19 @@ static LinkedHashMap<String, Object> secret(String secretName, String envVar) {
...
+      try {
+        sh """
+          git fetch origin demo:demo
+          git push --force origin HEAD:demo
+        """
+      } catch (err) {
+        // Send a notification to Slack channel?
```
timja-kainos: https://github.com/hmcts/moj-jenkins-library/blob/master/vars/notifyBuildEvent.groovy

Jenkinsfile_CNP
```diff
@@ -21,6 +21,19 @@ static LinkedHashMap<String, Object> secret(String secretName, String envVar) {
 }
 
 withPipeline("nodejs", product, component) {
+  afterCheckout {
+    onMaster {
+      try {
+        sh """
+          git fetch origin demo:demo
```
ckalista: Do we need to even fetch demo here?

rlewan: Yes, we need to. Jenkins wants to avoid unnecessary network traffic and fetches only the build branch and master, so it has no knowledge of demo until we fetch it explicitly.

geek-soft:
Cannot remember what US states but it is worth applying the same change to legal frontend.

Jenkinsfile_CNP
```diff
@@ -21,6 +21,19 @@ static LinkedHashMap<String, Object> secret(String secretName, String envVar) {
 }
 
 withPipeline("nodejs", product, component) {
+  afterCheckout {
```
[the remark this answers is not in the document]
geek-soft: Agree, we can always extract it to Jenkins library when other teams have the same need.

PR #580, opened 2018-05-11
Update to node 10

timja-kainos:


yarn.lock
```diff
@@ -79,7 +77,7 @@
 
 "@sinonjs/formatio@^2.0.0":
   version "2.0.0"
-  resolved "https://registry.yarnpkg.com/@sinonjs/formatio/-/formatio-2.0.0.tgz#84db7e9eb5531df18a8c5e0bfb6e449e55e654b2"
+  resolved "http://registry.npmjs.org/@sinonjs/formatio/-/formatio-2.0.0.tgz#84db7e9eb5531df18a8c5e0bfb6e449e55e654b2"
```
geek-soft: That is interesting. I wonder why yarn change domain.

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.1.1",
   "private": true,
   "engines": {
-    "node": ">=8.9.4 <10.0.0"
+    "node": ">=8.9.4"
```
sabahirfan: should this be ` >=10.1.0`?

timja-kainos: nah need jenkins to be upgraded for that, which requires all the teams to update

sabahirfan: ok

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:8.9.4-alpine
+FROM node:10.1.0-alpine
```
sabahirfan: The same need to be updated in `integration-tests.Dockerfile` as well

timja-kainos:
deferred for now

PR #581, opened 2018-05-14
Ronan civil procedure rules content

RonanFitzgerald:
image


src/main/views/terms-and-conditions.njk
```diff
@@ -15,6 +15,8 @@
         + 'This includes the ' + internalLink('privacy policy', privacyPolicyUri)) | safe }}
       </p>
 
+      <p>{{ t('By creating an account, you agree to accept documents from the court electronically.')</p>
```
filipczyks: }} missing?

timja-kainos: serves me right for fixing it in github 👅 
At least the routes test picked up on the issue

PR #583, opened 2018-05-15
ROC-3738: Do you need more time to respond? not available on due day

rlewan:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-3738

### Change description

We had a bug where the "Do you need more time to respond?" section was not available when defendant accessed the system on due day.

### Work checklist

- [ ] Yes
- [X] No

### DIFF
--- a/src/main/features/response/helpers/taskListBuilder.ts
+++ b/src/main/features/response/helpers/taskListBuilder.ts
@@ -2,7 +2,7 @@ import { TaskList } from 'drafts/tasks/taskList'
 import { TaskListItem } from 'drafts/tasks/taskListItem'
 import { Paths } from 'response/paths'
 import { ResponseDraft } from 'response/draft/responseDraft'
-import { Moment } from 'moment'
+import * as moment from 'moment'
 import { MomentFactory } from 'shared/momentFactory'
 import { MoreTimeNeededTask } from 'response/tasks/moreTimeNeededTask'
 import { OweMoneyTask } from 'response/tasks/oweMoneyTask'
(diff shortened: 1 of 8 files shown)


src/main/features/response/helpers/taskListBuilder.ts
```diff
@@ -25,7 +25,7 @@ export class TaskListBuilder {
     )
 
     const now: Moment = MomentFactory.currentDateTime()
```
geek-soft: Would using `MomentFactory.currentDate` method here work the same?

src/main/features/response/helpers/taskListBuilder.ts
```diff
@@ -25,7 +25,7 @@ export class TaskListBuilder {
...
-    if (claim.responseDeadline.isAfter(now)) {
+    if (claim.responseDeadline.endOf('day').isAfter(now)) {
```
geek-soft: On the second thought in the frontend we say:

`'You’ll have until 4pm on ' + claim.responseDeadline.format('LL') + ' to respond'`

Can you ask business whether we to use end of day or 4pm as deadline?

rlewan: Will do that now.

rlewan: It has to be 4 PM @geek-soft, very good spot. I also need to check if CCJ is working correctly with regard to this.

src/main/features/response/helpers/taskListBuilder.ts
```diff
@@ -11,9 +11,10 @@ import { YourDetails } from 'response/tasks/yourDetails'
 import { FreeMediationTask } from 'response/tasks/freeMediationTask'
 import { Claim } from 'claims/models/claim'
 import { WhenDidYouPayTask } from 'response/tasks/whenDidYouPayTask'
+import { isPastResponseDeadline } from 'claims/isPastResponseDeadline'
 
 export class TaskListBuilder {
-  static buildBeforeYouStartSection (draft: ResponseDraft, claim: Claim): TaskList {
+  static buildBeforeYouStartSection (draft: ResponseDraft, claim: Claim, now: moment.Moment): TaskList {
```
robertparkinson: why is this being passed in? Should it not just be set inside the method?

rlewan: Without it testing becomes very hard.

robertparkinson: fair enough

PR #588, opened 2018-05-16
Application Insights: obfuscate UUID's in url's which are not static files

filipczyks:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-3462

### Change description
Obfuscating UUID's in urls to {uuid} before being sent to App Insights

### Work checklist

- [ ] Yes
- [x] No

### DIFF
--- a/src/main/modules/app-insights/index.ts
+++ b/src/main/modules/app-insights/index.ts
@@ -6,11 +6,11 @@ declare class AppInsightsConfiguration {
   roleName: string
 }
 
-function hideUuidInUrlIfNotStaticFile (url: string): string {
-  let fileRegexp = new RegExp('(\\..{2,5}$)')
-  let uuidRegexp = new RegExp('[a-f0-9]{8}-[a-f0-9]{4}-4[a-f0-9]{3}-[89aAbB][a-f0-9]{3}-[a-f0-9]{12}')
+const fileRegexp = new RegExp('(\\..{2,5}$)')
+const uuidRegexp = new RegExp('[a-f0-9]{8}-[a-f0-9]{4}-4[a-f0-9]{3}-[89aAbB][a-f0-9]{3}-[a-f0-9]{12}')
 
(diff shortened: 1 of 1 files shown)


src/main/modules/app-insights/index.ts
```diff
@@ -6,13 +6,37 @@ declare class AppInsightsConfiguration {
   roleName: string
 }
 
+function hideUuidInUrlIfNotStaticFile (url: string): string {
+  let fileRegexp = new RegExp('(\\..{2,5}$)')
```
timja-kainos: what's this regexp for?

src/main/modules/app-insights/index.ts
```diff
@@ -6,13 +6,37 @@ declare class AppInsightsConfiguration {
...
+  let uuidRegexp = new RegExp('[a-f0-9]{8}-[a-f0-9]{4}-4[a-f0-9]{3}-[89aAbB][a-f0-9]{3}-[a-f0-9]{12}')
```
timja-kainos: this regex could be constructed once outside of the function and it should be a `const` not `let`

src/main/modules/app-insights/index.ts
```diff
@@ -6,13 +6,37 @@ declare class AppInsightsConfiguration {
...
+
+  if (url.match(fileRegexp) == null) {
```
timja-kainos: I think .test is more appropriate here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test

src/main/modules/app-insights/index.ts
filipczyks: Checking if URL is static file, ending with dot and 2 to 5 letters extension. So it prevents us not to rename files.

timja-kainos: O

PR #598, opened 2018-05-22
ROC-3655 enable and update full admission response to money claim page

rattur:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-3655

### Change description

Update existing components, routes, templates and test cases for enabling full admission in response

### Work checklist

- [ ] Yes
- [x] No


src/main/features/response/helpers/taskListBuilder.ts
```diff
@@ -75,13 +76,24 @@ export class TaskListBuilder {
       )
     }
 
+    if (draft.isResponseFullyAdmitted()) {
+      tasks.push(
+        new TaskListItem(
+          'Decide how you`ll pay?',
```
robertparkinson: there shouldn't be a ? here

src/main/features/response/helpers/taskListBuilder.ts
```diff
@@ -75,13 +76,24 @@ export class TaskListBuilder {
...
+          Paths.defencePaymentOptionsPage.evaluateUri({ externalId: externalId }),
+          WhenWillYouPayTask.isCompleted(draft)
```
robertparkinson: please refactor this to something else. maybe DecideHowYouWillPayTask ?

src/main/features/response/routes/when-will-you-pay.ts
```diff
@@ -25,7 +25,7 @@ function formLabelFor (responseDraft: ResponseDraft): string {
   if (isAmountTooHighPartialResponse(responseDraft)) {
     return 'When will you pay the amount you admit you owe?'
   } else {
-    return 'When will you pay?'
+    return 'How do you want to pay?'
```
robertparkinson: please refactor the name of the files and the link in the paths file too

src/main/features/response/views/when-will-you-pay.njk
```diff
@@ -24,10 +29,17 @@
           labelForScreenReadersOnly = true,
           name = 'option',
           options = [
+            {
+              label: DefendantPaymentType.IMMEDIATELY.displayValueFor(responseType),
+              value: DefendantPaymentType.IMMEDIATELY.value,
+              checked: (form.valueFor('option') === DefendantPaymentType.IMMEDIATELY),
+              hint: t('You must make sure they receive the money within 5 days. Any cheques or transfers must be clear in their account.')
```
robertparkinson: this is translated inside the macro. please remove the t()

src/main/features/response/views/when-will-you-pay.njk
```diff
@@ -14,8 +14,13 @@
...
         {{ csrfProtection(csrf) }}
 
+        <div class="notice">
+          <i class="icon icon-important"><span class="visually-hidden">{{ t('Warning') }}</span></i>
+          <strong class="bold-small">{{ t('You may receive a County Court Judgment (CCJ) if you don`t pay immediately.') }}</strong>
+        </div>
+
         {% if statementOfMeansIsApplicable %}
-          {% set hintText = t('You will need to give details of your finances to {{ claimantFullName }} to explain why.', { claimantFullName: claim.claimData.claimant.name }) %}
+          {% set hintText = t('You will need to give details of your finances to {{ claimantFullName }} to explain why you can`t pay immediately.', { claimantFullName: claim.claimData.claimant.name }) %}
```
robertparkinson: this is translated in the macro. Please remove the t()

src/main/features/response/routes/response-type.ts
```diff
@@ -55,7 +57,11 @@ export default express.Router()
             res.redirect(Paths.sendYourResponseByEmailPage.evaluateUri({ externalId: claim.externalId }))
             break
           case ResponseType.FULL_ADMISSION:
-            res.redirect(Paths.sendYourResponseByEmailPage.evaluateUri({ externalId: claim.externalId }))
+            if (toBoolean(config.get<boolean>('featureToggles.fullAdmission'))) {
```
timja-kainos: Please use the `FeatureToggles` class instead of using config directly

src/main/features/response/routes/decide-how-you-will-pay.ts
```diff
@@ -68,6 +68,9 @@ export default express.Router()
 
           const { externalId } = req.params
           switch (form.model.option) {
+            case DefendantPaymentType.IMMEDIATELY:
+              res.redirect(Paths.taskListPage.evaluateUri({ externalId: externalId }))
+              break
```
yarpo: I'd rather use `return` here. 

Why?
If you add anything after this switch-case it may be executed before you are redirected.

timja-kainos: you can do return res.redirect for a one-liner

src/integration-test/tests/citizen/defence/steps/defence.ts
```diff
@@ -284,39 +284,21 @@ export class DefenceSteps {
...
+        defendantWhenWillYouPage.chooseImmediately()
+        break
+
       default:
         throw new Error('Unknown DefenceType')
     }
 
-    this.checkAndSendAndSubmit(defendantType)
-    if (defenceType === DefenceType.FULL_REJECTION_WITH_DISPUTE || defenceType === DefenceType.FULL_REJECTION_BECAUSE_FULL_AMOUNT_IS_PAID) {
+    if (defenceType === DefenceType.FULL_REJECTION_WITH_DISPUTE
```
geek-soft: Why do we have this condition? Shouldn't all supported defence types end on check and send page?

src/main/features/response/form/models/defendantPaymentOption.ts
```diff
@@ -2,14 +2,16 @@ import { IsDefined, IsIn } from 'class-validator'
 import { ResponseType } from 'response/form/models/responseType'
 
 export class DefendantPaymentTypeLabels {
-  static readonly INSTALMENTS: string = 'By instalments'
+  static readonly INSTALMENTS: string = 'I`ll suggest a repayment plan'
   static readonly FULL_ADMIT_BY_SPECIFIED_DATE: string = 'Full amount by a set date'
   static readonly BY_SET_DATE: string = 'By a set date'
```
geek-soft: Why do we have both `FULL_ADMIT_BY_SPECIFIED_DATE` and `BY_SET_DATE`? It seems like we could use `FULL_ADMIT_BY_SPECIFIED_DATE`. @rattur please confirm and make appropriate removal.

src/main/features/response/routes/when-will-you-pay.ts
geek-soft: I would prefer naming path, route and related files after properties of the draft so it is easier to find how draft property is bound.

In that case it would be `.../response/defendant-payment-option` and `defendant-payment-option.ts`

src/main/features/response/routes/response-type.ts
geek-soft: It should work, how did you use it?

src/main/features/response/tasks/decideHowYouWillPayTask.ts
```diff
@@ -25,6 +25,8 @@ export class WhenWillYouPayTask {
         && this.explanationIsValidIfRequired(responseDraft.payBySetDate)
     } else if (responseDraft.defendantPaymentOption.option === DefendantPaymentType.INSTALMENTS) {
       return isValid(responseDraft.defendantPaymentPlan)
+    } else if (responseDraft.defendantPaymentOption.option === DefendantPaymentType.IMMEDIATELY) {
+      return isValid(responseDraft.defendantPaymentOption)
```
geek-soft: This will run validation second time. Please use switch statement and break early for `DefendantPaymentType.IMMEDIATELY` option.

src/main/features/response/views/when-will-you-pay.njk
geek-soft: Actually it makes sense to do it here, otherwise we would not be able to pass claimant name. This way interpolated and translated string will be passed for another translation that will result in the same string returned.

src/main/features/response/routes/when-will-you-pay.ts
rattur: Ok instead of using hardcoded values

src/main/features/response/views/when-will-you-pay.njk
rattur: Ok

src/main/features/response/routes/response-type.ts
rattur: FeatureToggleGuard.anyFeatureEnabledGuard('fullAdmission')

PR #604, opened 2018-05-24
Add offer accept guard when offer is already accepted

ellie-harrison:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-3764

### Change description
Added a route guard for accessing the offer acceptance declaration page when an offer has already been accepted.

### Work checklist

- [x ] I have clicked through the running application to see if all changes I made actually work.

- [ ] Yes
- [ x] No


src/main/features/offer/routes/declaration.ts
```diff
@@ -23,15 +24,15 @@ function renderView (form: Form<Declaration>, res: express.Response) {
 /* tslint:disable:no-default-export */
 export default express.Router()
   .get(
-    Paths.declarationPage.uri,
+    Paths.declarationPage.uri, OfferAcceptedGuard.requestHandler,
```
geek-soft: Because declaration is used for both singing an offer and counter signing I think current implementation would break counter signing.

@ellie-harrison can you make sure that counter signing works?

If not we would need to change the guard to mimic the following logic I found in the `declaration.ts`:

```
        if (user.id === claim.defendantId) {
          await OfferClient.countersignOffer(claim.externalId, user)

          res.redirect(Paths.settledPage.evaluateUri({ externalId: claim.externalId }))
        } else {
          await OfferClient.acceptOffer(claim.externalId, user)

          res.redirect(Paths.acceptedPage.evaluateUri({ externalId: claim.externalId }))
        }
```

geek-soft: @ellie-harrison any update on this?

ellie-harrison: This does affect countersigning, currently implementing a solution to stop that happening.

src/main/features/offer/guards/offerAcceptedGuard.ts
```diff
@@ -0,0 +1,26 @@
...
+import { Paths } from 'dashboard/paths'
+import { Logger } from '@hmcts/nodejs-logging'
+import { Claim } from 'claims/models/claim'
+import { User } from 'idam/user'
+
+const logger = Logger.getLogger('offer/guards/offerAcceptedGuard')
+
+export class OfferAcceptedGuard {
+
+  static requestHandler (req: express.Request, res: express.Response, next: express.NextFunction): void {
```
geek-soft: @ellie-harrison can you please use GuardFactory so that:
- we have consistent way of creating guards
- factory based way does not require any changes to a11y tests

Otherwise you would need to mocks your guard in a11y mocks.

See OnlyClaimantLinkedToClaimCanDoIt as an example.

PR #615, opened 2018-06-05
Make minor content changes

RonanFitzgerald:
1. I changed the dash, adding 'you said' to this line so it says 'telling them when and how you said you paid the claim' - the previous wording implies the service agrees their defence is true.
2. I changed 'your offer has been sent' to 'we've sent your offer' - just because our style is make it clear who has done something and not use the passive voice.

### Work checklist

- [ ] Yes
- [ ] No


src/main/features/dashboard/views/macro/claimStatus/offerSubmitted.njk
```diff
@@ -19,7 +19,7 @@
 
 {% macro offerSubmittedForDefendantDashboard() %}
   <p>
-    {{ t('Your offer has been sent to the claimant.') }}
+    {{ t('We've sent your offer to the claimant.') }}
```
geek-soft: Please use apostrophe char instead of single quote.

geek-soft:
Before merging we need to make sure that there are no other places with the same text that could be updated.

PR #621, opened 2018-06-14
ROC-2687: Add E2E tests for submitting full admission to claim-store

geek-soft:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-2687

### Change description

Changes handle submission full admission to claim-store.

Feature is build on top of the other feature branch does still needs to be merged into master so temporarily PR points towards other feature branch to narrow change set.

### Work checklist

- [ ] Yes
- [x] No


src/main/features/response/views/confirmation/full-defence-steps.njk
```diff
@@ -0,0 +1,18 @@
+{% set claimantName = claim.claimData.claimant.name %}
+
+<p>{{ t('If {{ claimantName }} accepts your defence the claim will be cancelled.', { claimantName: claimantName }) | safe }}</p>
+<p>{{ t('If they reject your defence the case will be reviewed and might go to a hearing.') }}</p>
+
+{% if featureToggles.offer %}
```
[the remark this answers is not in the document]
geek-soft: ROC-3701, been created some time ago, moved it to kanban live support board now.

src/main/app/claims/models/response/responseCommon.ts
```diff
@@ -1,30 +1,30 @@
...
-  FULL_DEFENCE = 'FULL_DEFENCE'
-}
+import { StatementOfTruth } from 'claims/models/statementOfTruth'
 
 export interface ResponseCommon {
   responseType: ResponseType
-  freeMediation: string
-  moreTimeNeeded?: string
   defendant: Party
+  moreTimeNeeded?: YesNoOption
```
timja-kainos: why a field rather than a boolean?

src/main/features/response/views/confirmation.njk
```diff
@@ -9,40 +10,41 @@
...
+        {% include "./confirmation/full-admission-steps.njk" %}
+      {% endif %}
+      {% if claim.response.responseType === domain.ResponseType.FULL_DEFENCE %}
+        {% include "./confirmation/full-defence-steps.njk" %}
       {% endif %}
-      {{ supportWidget() }}
 
       {% include "includes/service_survey_link.njk" %}
+
+      <p><a class="button" href="{{ DashboardPaths.dashboardPage.uri }}">{{ t('Go to your account') }}</a></p>
```
robertparkinson: I don't think this should be in a p tag as it's not really text/content. can you try just using form-group to see if you get the same effect?

src/main/features/response/views/confirmation/full-admission-steps.njk
```diff
@@ -0,0 +1,28 @@
+{% set claimantName = claim.claimData.claimant.name %}
+{% set claimantDetailsPageURI = ClaimPaths.claimantDetailsPage.evaluateUri({ externalId: claim.externalId }) %}
+
+{% if claim.response.paymentOption === domain.PaymentOption.IMMEDIATELY %}
+  <p>{{ t('When {{ claimantName }} gets your payment they must tell you and the court you’ve paid.', { claimantName: claimantName }) }}</p>
+{% endif %}
+
+{% if claim.response.paymentOption === domain.PaymentOption.FULL_BY_SPECIFIED_DATE %}
+  <p>{{ t('If {{ claimantName }} accepts your offer to pay, you need to <a href="{{ link }}">contact them for payment details</a> and to arrange payment.', { claimantName: claimantName, link: claimantDetailsPageURI }) | safe }}</p>
```
robertparkinson: should we be using internalLink macro for all these links?

src/main/features/response/views/confirmation/full-defence-steps.njk
```diff
@@ -0,0 +1,18 @@
...
+  <h2 class="heading-medium">{{ t('Settle out of court') }}</h2>
+  <p>{{ t('You may still <a href="{{ link }}">settle the claim out of court</a>. For example you could offer to fix goods you sold the claimant or suggest a payment.',
```
robertparkinson: more internalLink opportunities

src/main/features/response/views/confirmation/full-defence.njk
```diff
@@ -0,0 +1,2 @@
+<p>{{ t('We’ve emailed {{ claimant }} your defence, explaining why you reject the claim.', { claimant: claim.claimData.claimant.name }) }}</p>
+<p>{{ t('View the <a href="{{ link }}">claim against you and your response.</a>', { link: DashboardPaths.dashboardPage.uri }) | safe }}</p>
```
robertparkinson: more internalLink opportunities

src/main/app/claims/models/response/responseCommon.ts
geek-soft: Because it is like that on the API side :(

I will aim to remove `YesNoOption` from the claim-store model but that will be in the future. At least for now I am stopping people from using `YesNoOption` in CMC domain model ;)

timja-kainos: could use booleans here? and not let it creep in anymore?

geek-soft: I could define boolean and map it at deserialisation time but I personally do not see much benefit from doing it. With this when we remove YesNoOption from backend we would need to revisit that piece of code anyway.

geek-soft:
@robertparkinson I will address your comments here:  https://github.com/hmcts/cmc-citizen-frontend/pull/635

src/main/features/response/views/confirmation.njk
```diff
@@ -9,40 +10,41 @@
     <div class="column-two-thirds">
       <div class="govuk-box-highlight">
         <h1 class="bold-large">{{ t('You’ve submitted your response') }}</h1>
+        <p class="reference-number">
+          <span>{{ t('Your claim reference:') }}</span>
+          <span class="block bold-large">{{ claim.claimNumber }}</span>
```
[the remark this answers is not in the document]
geek-soft: I guess the reason we started doing it like that is to benefit from spacing coming from `p`.

We could use `from-group` but that is not ideal solution as there is not form element either.

src/main/features/response/views/confirmation.njk
geek-soft: We could use `from-group` but that is not ideal solution as there is not form element either.

Technically `<a>` is a text element in HTML despite being styled out as button by the CSS :)

src/main/features/response/views/confirmation/full-admission-steps.njk
geek-soft: I investigated use of `internalLink` and found that it has major flaw in terms of using it with i18n framework.

First issue is that because of `<a href="{{ href }}">{{ text }}</a>` URI gets interpolated at the macro level which means that `<a href="/some-uri">text</a>` is being passed to i18n framework. That forces us to use hardcoded links in *.po files which means more maintenance.

Second issue is that HTML printed by the macro has line break at the and which means that without trimming entries in *.po files would need to have the same line break :(

```
{% macro internalLink(text, href) %}
  <a href="{{ href }}">{{ text }}</a><-- here
{% endmacro %}
```

The same flaw occurs in `externalLink` macro :(

Due to the above I won't be using link macros in the PR.

src/main/app/claims/responseModelConverter.ts
```diff
@@ -111,10 +215,49 @@ export class ResponseModelConverter {
...
+  }
+
   private static convertWhenDidYouPay (whenDidYouPay: WhenDidYouPay): PaymentDeclaration {
     if (whenDidYouPay === undefined) {
       return undefined
     }
     return new PaymentDeclaration(whenDidYouPay.date.asString(), whenDidYouPay.text)
   }
+
+  private static convertStatementOfMeansChildreen (draft: ResponseDraft): Child[] {
```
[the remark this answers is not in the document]
geek-soft: Done https://github.com/hmcts/cmc-citizen-frontend/pull/635/commits/be0c63b88fb9bfbff2ff3b628fbccf4541537dd5

src/main/features/claim/views/claimant-details.njk
```diff
@@ -0,0 +1,51 @@
...
+        {{ claim.claimData.claimant.address.city }}<br />
+        {{ claim.claimData.claimant.address.postcode }}
+      </p>
+      {% if claim.claimData.claimant.mobilePhone %}
+        <p>{{ t('Phone: {{ phoneNumber }}', { phoneNumber: claim.claimData.claimant.mobilePhone }) }}</p>
+      {% endif %}
+      <p>{{ t('Email: <a href="mailto:{{ claimantEmail }}">{{ claimantEmail }}</a>', { claimantEmail: claim.claimantEmail }) | safe }} </p>
+    </div>
+
+    <div class="column-one-third">
```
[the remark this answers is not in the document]
geek-soft: Done https://github.com/hmcts/cmc-citizen-frontend/pull/635/commits/9586b399dff6bcea909213efa5b1fc6453e5d7c6

src/integration-test/tests/citizen/defence/full_admission_test.ts
```diff
@@ -0,0 +1,37 @@
...
+Before(function* (I: I) {
+  const claimantEmail: string = yield I.createCitizenUser()
+  const defendantEmail: string = yield I.createCitizenUser()
+
+  const claimData: ClaimData = createClaimData(PartyType.INDIVIDUAL, PartyType.INDIVIDUAL)
+  const claimRef: string = yield I.createClaim(claimData, claimantEmail)
+
+  yield helperSteps.enterPinNumber(claimRef, claimantEmail)
+  helperSteps.linkClaimToDefendant(defendantEmail)
+  helperSteps.startResponseFromDashboard(claimRef)
```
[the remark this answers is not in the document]
geek-soft: It seems they fixed the issue. Screenshot is taken and indicates the issue happened in the hook.

The only difference is that test execution stops when first issue occurred which means that if grep expression is picking for example 3 tests behind the same hook then failure in the hook of the first test case will stop other tests from being fired.

I guess the assumption here is that since all tests share the same prep logic if one failed there is hight risk that the other will fail as well.

I would be keep to leave it as is and see whether we can have any issues with it. Alternatively I can extract the logic into method and call that method in every tests.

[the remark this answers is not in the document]
geek-soft: screen shot 2018-06-25 at 17 46 59

geek-soft:
Due to the fact preview environments are broken for many days I am merging this on the basis that local build has succeded.

screen shot 2018-07-04 at 10 46 03

PR #647, opened 2018-07-02
Move to CNP fees and payments (non prod only)

timja-kainos:
Somehow their certs expired and no one renewed them, migrating to cnp early to resolve

### DIFF
--- a/infrastructure/prod.tfvars
+++ b/infrastructure/prod.tfvars
@@ -1,7 +1,7 @@
 vault_section = "prod"
 idam_api_url = "https://prod-idamapi.reform.hmcts.net:3511"
 authentication_web_url = "https://hmcts-access.service.gov.uk"
-fees_api_url = "https://fees-register.reform.hmcts.net:4411"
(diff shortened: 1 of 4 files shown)


infrastructure/prod.tfvars
```diff
@@ -7,4 +7,3 @@ ga_tracking_id = "UA-97111056-2"
 external_host_name = "www.moneyclaims.service.gov.uk"
 feature_return_error_to_user = "false"
 capacity = "2"
-
```
geek-soft: Can we use CNP fees API in prod as well?

pragnesh-patel:
Approved @timja-kainos but some checks were not successful

infrastructure/prod.tfvars
jaleen: yup, it's over https using certs which have CA errors at the moment which should be acceptable

timja-kainos:
@pragnesh-patel that's a unit test failure being worked on by ellie, its related to dates and it isn't working in july for some reason

PR #675, opened 2018-07-13
ROC-3803: Preventing double clicks

singh-surjeet:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-3803

### Change description

Preventing double clicks on citizen app. Will be replicated on legal once approved. 

- [ ] Yes
- [X] No


src/main/views/includes/scripts.njk
```diff
@@ -7,6 +7,7 @@
 <script src="{{ asset_paths['js'] }}/detectJavascript.js"></script>
 <script src="{{ asset_paths['js'] }}/analytics.js"></script>
 <script src="{{ asset_paths['js'] }}/analytics-custom-event-sender.js?2017-10-11"></script>
+<script src="{{ asset_paths['js'] }}/double-click-preventer.js"></script>
```
timja-kainos: when you add querySelectorAll you should add above it the nodelist polyfill,
`<script src="{{ asset_paths['js_vendor'] }}/nodelist-foreach-polyfill.js"></script>`

src/main/public/js/double-click-preventer.js
```diff
@@ -0,0 +1,14 @@
+(function () {
+  document.addEventListener('DOMContentLoaded', function (event) {
+
+    var submitType = document.querySelector('input[type=submit]')
+
+    if (submitType != null) {
+      submitType.addEventListener('dblclick', function (event) {
```
timja-kainos: While this story says preventing dblclick what we're really trying to achieve is not sending multiple requests to the backend, I think a click event would be better and then disabling the input

geek-soft: I agree. I can imagine that with backend responding slow user would be able to fire two separate clicks that are not handled by double click event. The ultimate goal is to prevent multiple same requests being fired to backend.

timja-kainos: how is it not covered? 
would be worth checking IE with the above solution, I would have thought that as long as you disable the input after the first click it would submit fine?

singh-surjeet: Thanks all for the inputs. In conversation with @robertparkinson and GDS pattern we tends to use javascript first before using JQuery (where we have to maintain different version due to different supported browsers). We also discussed disabling submit button which may not go down well with UX. Testing other review comments but we can check with UX on this.

src/main/public/js/double-click-preventer.js
```diff
@@ -0,0 +1,14 @@
...
+        event.preventDefault()
```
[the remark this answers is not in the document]
robertparkinson: Have had discussions with pete around this before and he doesn't like buttons showing as disabled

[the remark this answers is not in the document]
timja-kainos: Can we check with our gov pay works?
i.e. would be nice to be consistent and it does look good imo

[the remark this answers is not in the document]
singh-surjeet: Here's the javascript - https://github.com/alphagov/pay-frontend/blob/master/app/assets/javascripts/modules/stop-double-submission.js

src/main/public/js/double-click-preventer.js
```diff
@@ -0,0 +1,12 @@
+(function () {
+  document.addEventListener('DOMContentLoaded', function (event) {
+
+    var submitInputTypeList = document.querySelectorAll('input[type=submit]')
+      .forEach(function (item) {
+        item.addEventListener('dblclick', function (event) {
```
timja-kainos: this doesn't address the intent of the story which is stopping multiple clicks / a click then a pause then another one,
also have you shown how pay does it to UX?

src/main/public/js/double-click-preventer.js
```diff
@@ -0,0 +1,18 @@
+(function () {
+  document.addEventListener('DOMContentLoaded', function (event) {
+
+    var submitInputTypeList = document.querySelectorAll('input[type=submit]')
+      .forEach(function (element) {
+        element.addEventListener('click', function (event) {
+          if(element.className === 'button') {
```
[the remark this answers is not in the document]
robertparkinson: but not all objects with class 'button' are submit. there are a few a links that also have the button class and will need to test to make sure it doesn't effect it

src/main/public/js/double-click-preventer.js
```diff
@@ -0,0 +1,18 @@
+(function () {
+  document.addEventListener("DOMContentLoaded", function (event) {
+
+    var submitInputTypeList = document.querySelectorAll("input[type=submit]")
```
timja-kainos: this variable is never used

src/main/public/js/double-click-preventer.js
```diff
@@ -0,0 +1,18 @@
+(function () {
+  document.addEventListener("DOMContentLoaded", function (event) {
```
timja-kainos: although its not enforced for the frontend javascript our codestyle is to use `standardjs`
which means always use single quotes and no semi-colons

singh-surjeet: https://app.codacy.com happen to be moaning about double quotes and semi colon hence I decided to do it. Their argument on double quote felt quite reasonable to me as well.

timja-kainos: ignore codacy when it violates our coding style please

geek-soft:
Few questions before I run it

src/main/public/js/double-click-preventer.js
```diff
@@ -0,0 +1,18 @@
+(function () {
+  document.addEventListener('DOMContentLoaded', function (event) {
+
+    document.querySelectorAll('input[type=submit]')
+      .forEach(function (element) {
+        element.addEventListener('click', function (event) {
+          if(element.className === 'button') {
+            if(!event.detail || event.detail !== 1) {
```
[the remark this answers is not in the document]
singh-surjeet: I've tested on check-and-send which redirects to payments being the slowest page of the journey. Happy to test on other pages.

src/main/public/js/double-click-preventer.js
```diff
@@ -0,0 +1,18 @@
+(function () {
+  document.addEventListener('DOMContentLoaded', function (event) {
+
+    document.querySelectorAll('input[type=submit]')
+      .forEach(function (element) {
+        element.addEventListener('click', function (event) {
+          if(element.className === 'button') {
```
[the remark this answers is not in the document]
singh-surjeet: E.g. <input class="button button-secondary" name="action[addRow]" ..> when additional rows are added which is quite commonly used.

singh-surjeet:
Tested it on chrome where chrome Dev tool simulates to poor 3G connection.

src/main/public/js/double-click-preventer.js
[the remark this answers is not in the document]
timja-kainos: why do  we need to check the submit button's class?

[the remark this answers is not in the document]
timja-kainos: ok, missing space `if` before the bracket

src/main/public/js/double-click-preventer.js
```diff
@@ -0,0 +1,17 @@
+(function () {
+  document.addEventListener('DOMContentLoaded', function (event) {
+
+    document.querySelectorAll('input[type=submit]')
+      .forEach(function (element) {
+        element.addEventListener('click', function (event) {
+          if (element.className === 'button') {
```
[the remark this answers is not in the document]
singh-surjeet: Correct. UX wish to apply disabling where necessary.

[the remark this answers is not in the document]
geek-soft: Probably only matters to postcode lookup and monthly income and expense pages. I guess we can leave it like it is.

src/main/public/js/double-click-preventer.js
```diff
@@ -0,0 +1,17 @@
...
+            element.disabled = true
+            element.form.submit()
```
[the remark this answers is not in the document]
geek-soft: Interesting that disabling element stops event propagation.

src/main/public/js/double-click-preventer.js
```diff
@@ -0,0 +1,17 @@
+(function () {
+  document.addEventListener('DOMContentLoaded', function (event) {
```
geek-soft: Remove unused parameter.

src/main/public/js/double-click-preventer.js
```diff
@@ -0,0 +1,17 @@
...
+
+    document.querySelectorAll('input[type=submit]')
+      .forEach(function (element) {
+        element.addEventListener('click', function (event) {
```
geek-soft: Remove unused parameter.

src/main/public/js/double-click-preventer.js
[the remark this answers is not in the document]
singh-surjeet: If we disable button like we're doing before submit() then we needed to submit ourselves for submit input in question otherwise it won't submit.

src/main/public/js/double-click-preventer.js
singh-surjeet: Removed.

geek-soft: Not removed. Only 50% of what I asked got removed.

singh-surjeet: Ohh yes, sorry!. Removed

PR #682, opened 2018-07-18
ROC-3859 Reset income / expense fix

timja-kainos:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-3859

### Change description
Fix reset for JS off

### Work checklist

- [ ] Yes
- [x] No

### DIFF
--- a/src/main/features/response/form/models/statement-of-means/monthlyIncome.ts
+++ b/src/main/features/response/form/models/statement-of-means/monthlyIncome.ts
@@ -172,4 +172,9 @@ export class MonthlyIncome {
     this.otherSources.splice(this.otherSources.findIndex(element => element === source), 1)
   }
 
+  resetIncome (propertyName: string, source: IncomeSource): void {
+    this[`${propertyName.split('.')[0]}Declared`] = false
+    source.reset()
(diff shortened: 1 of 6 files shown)


src/main/features/response/form/models/statement-of-means/monthlyIncome.ts
```diff
@@ -172,4 +172,9 @@ export class MonthlyIncome {
     this.otherSources.splice(this.otherSources.findIndex(element => element === source), 1)
   }
 
+  resetIncome (propertyName: string, source: IncomeSource): void {
+    this[`${propertyName.split('.')[0]}Declared`] = false
```
ckalista: Fully understand why we have this here. Maybe a bit much responsibility for the model. Wonder if there would be an easy enough way to doing the `propertyName` manipulation outside of the model?

timja-kainos: I take your point but I think it's ok here, its encapsulated in resetting a field, which the model should know best how to do.

unfortunately there's the complexity of there being two fields that need clearing

PR #684, opened 2018-07-19
ROC-2703 Refactor api data model change in partial admission confirma…

dharmendrak:


config/default.json
```diff
@@ -84,10 +84,10 @@
     "testingSupport": false,
     "countyCourtJudgment": false,
     "fullAdmission": false,
-    "partialAdmission": false,
-    "admissions": false,
+    "partialAdmission": true,
```
sabahirfan: I think these needs to be turned off by default. please revert this file

PR #689, opened 2018-07-20
ROC-4052: Add `a11y` tests for claimant response feature

sabahirfan:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-4052

### Change description
This PR adds `a11y` tests for claimant-response feature.

- [ ] Yes
- [x] No

### DIFF
--- a/src/test/a11y/a11y.ts
+++ b/src/test/a11y/a11y.ts
@@ -12,6 +12,7 @@ import { Paths as EligibilityPaths } from 'eligibility/paths'
 import { ErrorPaths as ClaimIssueErrorPaths, Paths as ClaimIssuePaths } from 'claim/paths'
 import { ErrorPaths as DefendantFirstContactErrorPaths, Paths as DefendantFirstContactPaths } from 'first-contact/paths'
 import { Paths as DefendantResponsePaths, StatementOfMeansPaths, FullAdmissionPaths } from 'response/paths'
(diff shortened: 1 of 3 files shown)


src/test/a11y/a11y.ts
```diff
@@ -92,7 +93,8 @@ const excludedPaths: DefendantResponsePaths[] = [
   DefendantResponsePaths.receiptReceiver,
   DefendantResponsePaths.legacyDashboardRedirect,
   OfferPaths.agreementReceiver,
-  DefendantFirstContactPaths.receiptReceiver
+  DefendantFirstContactPaths.receiptReceiver,
+  ClaimantResponsePaths.checkAndSendPage
```
geek-soft: Why check-and-send is excluded?

sabahirfan: not implemented yet.

geek-soft: We just need to remember to revert it when it is implemented so we are checking it as a part of a11y tests.

timja-kainos: can we not exclude it please? that's something easily forgotten

sabahirfan: don't worry, it won't be forgotten, at this point it would be un necessary work-around to to include and make it pass.

PR #707, opened 2018-07-25
[PREVIEW] ROC-4038: View defendants response

rattur:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-4038

### Change description

Currently only works for Full Admission, pay installments.  Needs additional checks for Full Admission on a set date and part admission. Work in progress to look at perhaps moving the logic from the route into the template

### Work checklist

- [ ] Yes
- [x] No


src/test/features/claimant-response/helpers/taskListBuilder.ts
```diff
@@ -24,13 +24,14 @@ describe('Claimant response task list builder', () => {
   let claimantResponseDraft: DraftClaimantResponse
 
   beforeEach(() => {
-    claim = new Claim().deserialize({ ...claimStoreServiceMock.sampleClaimObj, ...claimStoreServiceMock.sampleDefendantResponseObj })
+    claim = new Claim().deserialize({ ...claimStoreServiceMock.sampleClaimObj, ...claimStoreServiceMock.sampleDefendantResponseObj, ...claimStoreServiceMock })
```
sabahirfan: do we need `...claimStoreServiceMock` ? I think otherwise

src/main/app/common/calculate-monthly-income-expense/incomeExpenseSource.ts
```diff
@@ -57,6 +59,32 @@ export class IncomeExpenseSource {
       toIncomeExpenseScheduleOrUndefined(schedule)
     )
   }
+
+  static fromClaimIncome (income: Income): IncomeExpenseSource {
+    if (!income) {
+      return undefined
+    }
+
+    const schedule: string = income.frequency ? income.frequency : undefined
```
sabahirfan: do we need a conditional operator here? It would be same otherwise

src/main/app/common/calculate-monthly-income-expense/incomeExpenseSource.ts
```diff
@@ -57,6 +59,32 @@ export class IncomeExpenseSource {
...
+      toIncomeExpenseScheduleOrUndefined(schedule)
+    )
+  }
+
+  static fromClaimExpense (expense: Expense): IncomeExpenseSource {
+    if (!expense) {
+      return undefined
+    }
+
+    const schedule: string = expense.frequency ? expense.frequency : undefined
```
sabahirfan: do we need a conditional operator here? It would be same otherwise

src/main/features/claimant-response/views/defendants-response.njk
```diff
@@ -0,0 +1,167 @@
...
+
+          {{ financialDetailsBlock(totalMonthlyIncome, totalMonthlyExpenses, claim.response.statementOfMeans) }}
+        {% endif %}
+      {% endif %}
+
+      <form novalidate method="post">
+        {{ csrfProtection(csrf) }}
+
+        {% if claim.response.responseType === ResponseType.PART_ADMISSION.value and page !== 1 %}
+          <input type="hidden" name="action[showPage]" value="{{ page + 1 }}">
```
sabahirfan: I prefer this routing logic be moved out of view. We can have two pages with two routes and decide accordingly whether we just need to display one or both. Makes it easier to test and will reduce the code complexity. I am still trying to find how this logic is covered in tests.

src/main/app/common/calculate-monthly-income-expense/incomeExpenseSource.ts
geek-soft: Good spot. Removed.

src/main/app/common/calculate-monthly-income-expense/incomeExpenseSource.ts
geek-soft: Good spot. Removed.

src/test/features/claimant-response/helpers/taskListBuilder.ts
geek-soft: Definitely not, that was stupid, I have no idea how that made the code and how that disappeared. Magic.

src/main/views/macro/financialDetails.njk
```diff
@@ -0,0 +1,188 @@
...
+    <div class="column-one-half {% if not value %}if-empty{% endif %}">
+      <span>{{ value }}</span>
+    </div>
+  </div>
+{% endmacro %}
+
+{% macro bankAccountsSection(bankAccounts) %}
+  {% for bankAccount in bankAccounts %}
+    {% if bankAccounts.length === 1 %}
+      {{ row(label = 'Type of account', value = bankAccount.type | renderBankAccountType) }}
```
timja: are these filters useful?
i.e. don't most of these models have `displayValue` on them?

geek-soft: The approach with `displayValue` is just temporary. It will remain there until we remove these methods from TS classes. The reasoning behind using filters instead of constants with `displayValue` is that we want to separate UI from data.

src/main/features/claimant-response/views/defendants-response.njk
geek-soft: I looked into alternative of having separate route for details but it comes with lots of extra problems like:
 - name for the extra page
 - lots of additional code that we usually put in routes
 - necessity of sharing view macros between two pages
 - necessity of writing additional tests for extra route

In the meantime I added route test case for pagination.

PR #745, opened 2018-08-01
[PREVIEW] ROC-4049:Claimant response free mediation

varma02:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-4049

### Change description
This is first pull request for this story covering just free mediation


src/main/features/claimant-response/index.ts
```diff
@@ -44,7 +44,11 @@ export class ClaimantResponseFeature {
     app.all(/^\/case\/.+\/claimant-response\/(?!confirmation).*$/,
       DraftMiddleware.requestHandler(new DraftService(), 'claimantResponse', 100, (value: any): DraftClaimantResponse => {
         return new DraftClaimantResponse().deserialize(value)
-      }))
+      }),
+      (req: express.Request, res: express.Response, next: express.NextFunction) => {
+        res.locals.draft = res.locals.claimantResponseDraft
```
geek-soft: Alias so that we do not have to specify what draft we deal with in components.

In the future we could change middleware to populate draft local env everywhere as we always have one draft per journey.

PR #794, opened 2018-08-09
ROC-3986 remove other toggle features

dharmendrak:


src/main/features/response/components/payment-intention/payment-option.ts
```diff
@@ -73,7 +73,7 @@ export class PaymentOptionPage {
 
   private renderView (form: Form<DefendantPaymentOption>, res: express.Response) {
     function isApplicableFor (draft: ResponseDraft): boolean {
-      if (!FeatureToggles.isEnabled('statementOfMeans')) {
+      if (!FeatureToggles.hasAnyAuthorisedFeature(res.locals.claim.features, 'admissions')) {
```
timja: extract claim to a var for type safety

src/main/features/response/guards/statementOfMeansStateGuard.ts
```diff
@@ -19,7 +19,7 @@ export class StatementOfMeansStateGuard {
     return GuardFactory.create((res: express.Response) => {
       const draft: Draft<ResponseDraft> = res.locals.responseDraft
 
-      return StatementOfMeansFeature.isApplicableFor(draft.document)
+      return StatementOfMeansFeature.isApplicableFor(res.locals.claim.features, draft.document)
```
timja: extract here too?

src/main/features/response/components/payment-intention/payment-option.ts
```diff
@@ -72,16 +73,17 @@ export class PaymentOptionPage {
...
   private renderView (form: Form<DefendantPaymentOption>, res: express.Response) {
+    const claim: Claim = res.locals.claim
+
     function isApplicableFor (draft: ResponseDraft): boolean {
-      if (!FeatureToggles.isEnabled('statementOfMeans')) {
+      if (!FeatureToggles.hasAnyAuthorisedFeature(claim.features, 'admissions')) {
         return false
       }
-      return draft.isResponseFullyAdmitted()
+      return draft.isResponseFullyAdmitted(claim.features)
```
geek-soft: I do not like the fact that now we have to pass features to draft methods. IMO we should separate logic of checking whether feature is enabled from logic checking state of the draft.

I would suggest:

```
# new helper method in separate file
areAdmissionsEnabled(claim: Claim): boolean {
  return FeatureToggles.hasAnyAuthorisedFeature(claim.features, 'admissions')
}

// in many places
ClaimFeatureToggles.areAdmissionsEnabled(claim) && draft.isResponseFullyAdmitted()

// or is other places
if (ClaimFeatureToggles.areAdmissionsEnabled(claim)) {
  if (draft.isResponseFullyAdmitted()) {
    // some logic
  }
}
```

src/main/features/testing-support/routes/create-claim-draft.ts
```diff
@@ -29,6 +32,8 @@ export default express.Router()
       draft.document = new DraftClaim().deserialize(prepareClaimDraft(user.email))
       await new DraftService().save(draft, user.bearerToken)
 
+      await claimStoreClient.addRoleToUser(user, 'cmc-new-features-consent-given')
```
timja: This doesn't check if the user has already given consent or withheld their consent?
I.e. a user who says no to consent should still be able to use this endpoint and not be opted in

src/main/features/testing-support/routes/create-claim-draft.ts
```diff
@@ -32,7 +32,11 @@ export default express.Router()
       draft.document = new DraftClaim().deserialize(prepareClaimDraft(user.email))
       await new DraftService().save(draft, user.bearerToken)
 
-      await claimStoreClient.addRoleToUser(user, 'cmc-new-features-consent-given')
+      const userRoles: string[] = await claimStoreClient.retrieveUserRoles(user)
+
+      if (userRoles.length === 0 || !userRoles.includes('cmc-new-features-consent-given')) {
```
timja: this check isn't right, as it doesn't stop you adding the role if its consent not given
```
!roles.some(role => role.includes('cmc-new-features-consent'))
```

PR #795, opened 2018-08-09
[PREVIEW] ROC-3986: Add new features header in claim store create method

dharmendrak:


src/integration-test/helpers/clients/claimStoreClient.ts
```diff
@@ -31,6 +31,7 @@ export class ClaimStoreClient {
    *
    * @param {ClaimData} claimData - claim data
    * @param {User} submitter - user that submits claim
+   * @headers Authorization, Features
```
timja: does this need to be here?

dharmendrak: no, i added for another commit

PR #837, opened 2018-08-21
Change data so less identifiable information in create claim draft - …

pragnesh-patel:
Need to get this in before Automation review


src/main/app/drafts/draft-data/claimDraft.ts
```diff
@@ -11,37 +11,37 @@ export function prepareClaimDraft (userEmailAddress: string) {
...
-            day: 27
+            year: 2000,
+            month: 1,
+            day: 1
           }
         }
       },
       mobilePhone: {
-        number: '07112358132'
+        number: '(0)207 127 0000'
```
timja: isn't this not a valid mobile?

jacmurphy: I asked this to pragnesh at the time and he said it was fine

timja: has it been tested? I don't think it is fine

pragnesh-patel: @timja @jacmurphy - Its asking a claimant for a telephone number - The field is free text and was changed to that, so I think its valid to add a telephone number -  I can refactor it, but I just did the minimum change here

timja: actually you're right 👍

PR #844, opened 2018-08-21
ROC-4289 Fix company CYA

timja:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-4289

### Change description
Add additional SoM page not displayed in SoM section of CYA 😠  to check for company and orgs

### Work checklist

- [ ] Yes
- [ ] No

### DIFF
--- a/src/main/features/response/views/macro/check-and-send/admission.njk
+++ b/src/main/features/response/views/macro/check-and-send/admission.njk
@@ -11,7 +11,7 @@
   }}
 {% endmacro %}
 
-{% macro notImmediatePayment(type, paymentDate, paymentPlan, explanation, Paths = {}, externalId) %}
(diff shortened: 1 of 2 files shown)


src/main/features/response/views/macro/check-and-send/admission.njk
```diff
@@ -11,7 +11,7 @@
   }}
 {% endmacro %}
 
-{% macro notImmediatePayment(type, paymentDate, paymentPlan, explanation, Paths = {}, externalId) %}
+{% macro notImmediatePayment(type, paymentDate, paymentPlan, explanation, Paths = {}, externalId, statementOfMeansIsApplicable) %}
```
geek-soft: I would prefer to pass undefined `explanation` if statement of means is not applicable.
Then in the macro we would have `Why you can’t pay now` section optional.

timja: I would prefer to keep the check the same for SoM features, its just a shame that the SoM task is in there too

PR #875, opened 2018-08-29
Update to new IDAM URL

robertparkinson:


geek-soft:
PR is misleading. API is the same, just domain is changed.

PR #951, opened 2018-09-21
ROC-4396: changes to statement of means calculation

tharacka:
https://tools.hmcts.net/jira/browse/ROC-4396

Prototypes:
https://tools.hmcts.net/jira/browse/ROC-4388


config/default.json
```diff
@@ -1,4 +1,5 @@
 {
```
andrewwa-kainos: Note: these are made-up values

PR #954, opened 2018-09-23
ROC-4322:Completion date in payment plan

varma02:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-4322

### Change description
Add completion date to payment plan


src/test/common/components/payment-intention/model/paymentPlan.ts
```diff
@@ -14,6 +14,7 @@ const DEFAULT_PAYMENT_PLAN = {
   instalmentAmount: 50,
   firstPaymentDate: { year: FUTURE_YEAR, month: 10, day: 10 },
   paymentSchedule: PaymentSchedule.EVERY_MONTH.value,
+  completionDate: MomentFactory.parse('2028-10-10'),
```
ckalista: Here you are only allowed to have primitive types. Remember the payload you are constructing is meant to simulate a form object sent by the browser.

src/test/common/components/payment-intention/model/paymentPlan.ts
```diff
@@ -22,11 +23,12 @@ const DEFENDANT_PAYMENT_PLAN_FOR_DESERIALISATION = {
   instalmentAmount: 50,
   firstPaymentDate: { year: FUTURE_YEAR, month: 10, day: 10 },
   paymentSchedule: { value: PaymentSchedule.EVERY_MONTH.value, displayValue: PaymentSchedule.EVERY_MONTH.displayValue },
+  completionDate: MomentFactory.parse('2028-10-10'),
```
ckalista: Same comment as above expect that the payload simulate an object returned from the draft store.

ckalista:
In addition to this, don't forget to also update the converter. Indeed the `last payment data` now has to be submitted to the claim-store then in the claim response journey everywhere we need it it should come the the claim object instead of being recalculated. Whole of this needs to be tested.

varma02:
> In addition to this, don't forget to also update the converter. Indeed the `last payment data` now has to be submitted to the claim-store then as part of the claim response journey everywhere we need it it should come the the claim object instead of being recalculated. Whole of this needs to be tested.

@ckalista - I understand what you are saying here but strictly speaking we don't need completion date in payment plan as part of defendant journey because we don't use it until claim response journey. In my other PR i am saving completion date in draft as part of claimant response jouney which can be used in convertor to store in claim store. 

Having said all this, i think it makes sense to save completion date as soon as user clicks on continue in payment plan pages (where ever that is). 

So, I don't think this should block anything if my other PR goes in. This will be an improvement to my other PR in my opinion. 

We will discuss tomorrow.

varma02:
different branch opened

PR #963, opened 2018-09-25
refactor to run bootstrap just once

markry-kainos:
### Change description

@robertparkinson reported issue running integration tests locally: `anyone been able to get the integration tests running locally recently? I keep getting the following when trying to run them:
Error during bootstrap, exiting { error: null }`

@timja suggested not running the bootstrap script 3 times. Makes perfect sense :) 

I decoupled bootstrap from codecept config (as we config to run 3 in parallel) and refactored to run before integration tests. Minimal code changes.

### Work checklist

- [X] I have run integration tests locally.

- [ ] Yes
- [X] No


src/integration-test/bootstrap.js
```diff
@@ -0,0 +1,3 @@
+process.env.TS_NODE_TRANSPILE_ONLY = "true"
```
timja: you could probably get rid of this file by using ts-node instead of node for running the bootstrap file

default.conf.js
```diff
@@ -5,7 +5,6 @@ const ProxySettings = require('./src/integration-test/config/proxy-settings').Pr
 
 exports.config = {
   name: 'citizen-integration-tests',
-  bootstrap: './src/integration-test/bootstrap/bootstrap.ts',
```
geek-soft: There must be easier way of doing it. I started looking into `bootstrapAll` which is executed only once per run regardless how many workers are there.

PR #1020, opened 2018-10-11
ROC-4451: Simplify completion date and payment length code

varma02:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-4451

### Change description
Simplify completion date and payment length code


varma02:
I will see if i can add more tests but please review everything else

PR #1021, opened 2018-10-11
ROC-4494 Set default ccj to be of type DEFAULT'

dharmendrak:


pragnesh-patel:
prob need to up the test coverage @dharmendrak for codecov to pass

PR #1022, opened 2018-10-11
ROC-4451:Simplified creating payment plan logic 

varma02:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-4451

### Change description
Simplified creating payment plan logic


src/main/app/claims/models/response/core/paymentIntention.ts
```diff
@@ -21,7 +21,7 @@ export class PaymentIntention {
       instalmentAmount: input.repaymentPlan.instalmentAmount,
       firstPaymentDate: MomentFactory.parse(input.repaymentPlan.firstPaymentDate),
```
tharacka: should the firstPaymentDate be subject to the same test as completionDate?

PR #1083, opened 2018-10-31
run only non-admissions e2e tests

markry-kainos:
### Change description

Stop admissions tests running on master

### Work checklist

- [ ] Yes
- [ ] No


bin/run-functional-tests.sh
```diff
@@ -21,6 +21,7 @@ if [[ "${1}" != "--no-build" ]]; then
   docker-compose -f ${ADDITIONAL_COMPOSE_FILE} build citizen-integration-tests
 fi
 docker-compose -f ${ADDITIONAL_COMPOSE_FILE} up --no-color -d remote-webdriver
-docker-compose -f ${ADDITIONAL_COMPOSE_FILE} run -u `id -u $USER` citizen-integration-tests
+# run non-admissions tagged tests ONLY
+docker-compose -f ${ADDITIONAL_COMPOSE_FILE} run -u `id -u $USER` citizen-integration-tests --grep '(?=.*)^(?!.*@admissions)'
```
markry-kainos: @timja can I pick your brain please :) we want to run tests when feature toggle=on and tests when feature toggle=off. I'm thinking we control here: so after this line: curl feature-toggle-api to switch toggle on then run relevant tests again via another docker-compose command. This sound sane to you?

markry-kainos: also for anyone looking the grep was copied from https://codecept.io/advanced/ :)

timja: How about another approach?
You should have tests that test when the toggle is enabled or disabled and can be run without changing global state

You can probably do this in the claim store via api for each case, as feature toggle api isn't a run time action other than saying whether the user should get the toggle or not.
i.e. via api call enable / disable admissions for a user.

basically this should be done at the user / case level and shouldn't need global toggling

markry-kainos: It makes sense... we have our AKS env (toggled on) and AAT (off)....admissions tests wont work on AAT - so our tests will test: if toggled on then x, if toggled off then y. Also we test - if toggled on and user doesn't have feature then x etc. Switching the user via claim-store api. Cool, think I'm on same page.

PR #1109, opened 2018-11-05
ROC-4679 Enable admissions e2e - depending on feature toggle

markry-kainos:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-4679

### Change description

Enabling our admissions e2e tests where feature toggle is on.

- [ ] Yes
- [ X ] No


Jenkinsfile_CNP
```diff
@@ -27,9 +27,13 @@ static LinkedHashMap<String, Object> secret(String secretName, String envVar) {
   ]
 }
 
-// Vars needed for AKS testing
+// Vars needed for AKS testing - keep values in sync with: /src/kubernetes/deployment.template.yaml
+// THESE WILL BE OVERWRITTEN IN ASE DEPLOYMENTS WITH TERRAFORM OUTPUTS.TF
+// THESE VARS CAN BE PICKED UP IN INTEGRATION-TEST CODE DIRECTLY
 env.IDAM_URL = 'https://preprod-idamapi.reform.hmcts.net:3511'
 env.CLAIM_STORE_URL = 'http://cmc-claim-store-aat.service.core-compute-aat.internal'
+env.FEATURE_ADMISSIONS = 'true'  // used to selective determine integration tests
```
markry-kainos: @timja ive been battling with this a bit...this is approach im happy with! AKS - env variables used in integration tests defined here, ASE defined in output.tf. Please don't spend any time looking at the actual e2e tests - we are due a major refactoring!!!

src/integration-test/tests/citizen/claimantResponse/full_admission_test.ts
```diff
@@ -15,122 +15,127 @@ const claimantResponseSteps: ClaimantResponseSteps = new ClaimantResponseSteps()
 const checkAndSendPage: ClaimantCheckAndSendPage = new ClaimantCheckAndSendPage()
 const confirmationPage: ClaimantConfirmation = new ClaimantConfirmation()
 
-Feature('Claimant Response').retry(3)
+if (process.env.FEATURE_ADMISSIONS !== undefined && JSON.parse(process.env.FEATURE_ADMISSIONS)) {
```
markry-kainos: @timja this line is of interest too... enabling if toggled

infrastructure/output.tf
```diff
@@ -13,3 +13,27 @@ output "idam_url" {
 output "claim_store_url" {
   value = "${local.claimStoreUrl}"
 }
+
+output "feature_admissions" {
```
timja: another way that's more full proof i.e. doesn't rely on people adding new ones here is to hit the /info endpoint, it has all feature toggles there

src/integration-test/tests/citizen/claimantResponse/full_admission_test.ts
timja: JSON.parse? isn't it just a string?

infrastructure/output.tf
markry-kainos: oh yes!!! yep - much better 👍

src/integration-test/tests/citizen/claimantResponse/full_admission_test.ts
markry-kainos: you would think but I had to do this to work locally (docker-compose) and on pipeline...maybe not - i will revisit this before any merge

markry-kainos:
@robertparkinson pls review - i think there is definitely a more codecept.io way to do this - using hooks (`event.all.before`) to detect toggle and set global variable. I couldn’t get it working! We can include this as part of refactoring work on the e2e

markry-kainos:
thanks @singh-surjeet - I added README as agreed: https://github.com/hmcts/cmc-citizen-frontend/pull/1109/files#diff-dce70a764e9461067cab9c77ebc6aa1a - will merge when I get green build 👍

PR #1131, opened 2018-11-08
ROC-4700:Get defendant paymentPlan only when defendant chooses instalment

varma02:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-4700

### Change description
Get defendant paymentPlan only when defendant chooses instalment


varma02:
Working on tests but want to share with others while in progress...

PR #1132, opened 2018-11-08
Remove redundant code for settlement endpoint

sabahirfan:
This PR removed the redundant call. We are not using this end point and being also remove in `claim-store-api` as well.


ckalista:
@sabahirfan Only get a chance to look into this now. Thanks a lot for the cleaning up 👍

PR #1153, opened 2018-11-15
ROC-4724: Handle negative disposable income to create payment intention of max future date

dharmendrak:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-4724

### Change description
Handle negative disposable income to create payment intention of max future date


src/main/app/common/court-calculations/courtOrder.ts
```diff
@@ -1,20 +0,0 @@
-export class CourtOrder {
```
ckalista: Thanks for removing this 👍

src/test/app/common/court-calculations/courtDecision.ts
```diff
@@ -43,11 +44,10 @@ describe('CourtDecision', () => {
       }).to.throw(Error, 'Input should be a moment, cannot be empty')
     })
 
-    it('should return a defendant decision type when no court date is supplied', () => {
-
-      let courtGeneratedPaymentDate = undefined
-
-      expect(CourtDecision.calculateDecision(null, null, courtGeneratedPaymentDate)).to.equal(DecisionType.DEFENDANT)
+    it('should return a defendant decision type when max future court date is supplied', () => {
```
ckalista: We are missing one important test, when court date is max future and claimant has made a more favourable offer than defendant. 

Or expressed in a different way, scenario [A] on the our court decision rules diagram should be tested with a court date that is max future and non max future.

ckalista:
Heya, just wondering if this covers the following as well or is it going to be dealt with as part of a separate PR?

- Negative disposable income should not be converted into zero but should be persisted as such in the draft and claimant response instead.
- `courtPaymentIntention` are populated when disposable incomes are negative.

src/test/app/common/court-calculations/courtDecision.ts
dharmendrak: added a test for it, other scenarios are all covered

src/integration-test/tests/citizen/claimantResponse/full_admission_test.ts
```diff
@@ -45,12 +45,10 @@ if (process.env.FEATURE_ADMISSIONS === 'true') {
...
-    // confirmationPage.clickGoToYourAccount()
-    // I.see(testData.claimRef)
-    // I.see('You’ve signed a settlement agreement')
+    I.see('You’ve accepted the repayment plan')
+    confirmationPage.clickGoToYourAccount()
+    I.see(testData.claimRef)
+    I.see('You’ve signed a settlement agreement')
   })
 
   Scenario('I can as a claimant accept the defendants full admission by set date with settlement agreement and rejecting defendants payment method in favour of immediate payment @citizen @admissions @error', async (I: I) => {
```
markry-kainos: remove @error tags now it is fixed :) also in other scenarios too

PR #1174, opened 2018-11-23
Disable cache in demo and prod

markry-kainos:
### Change description

Disabling cache for some devops testing by Ben Neil on recommendation by Microsoft. After outages 22 Nov.

https://github.com/hmcts/cnp-module-webapp/blob/master/variables.tf

- [ X ] Yes - POSSIBLY!!!
- [ ] No

### DIFF
--- a/infrastructure/prod.tfvars
+++ b/infrastructure/prod.tfvars
@@ -6,4 +6,4 @@ ga_tracking_id = "UA-97111056-2"
 external_host_name = "www.moneyclaims.service.gov.uk"
 feature_return_error_to_user = "false"
 capacity = "2"
-website_local_cache_sizeinmb = 0
+website_local_cache_sizeinmb = "0"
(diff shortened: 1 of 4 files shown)


infrastructure/prod.tfvars
```diff
@@ -6,3 +6,4 @@ ga_tracking_id = "UA-97111056-2"
 external_host_name = "www.moneyclaims.service.gov.uk"
 feature_return_error_to_user = "false"
 capacity = "2"
+website_local_cache_sizeinmb = 0
```
timja: should be a string

infrastructure/variables.tf
```diff
@@ -99,3 +99,7 @@ variable "capacity" {
 variable "common_tags" {
   type = "map"
 }
+
+variable "website_local_cache_sizeinmb" {
```
timja: could you add a comment about why you're overriding the default module-webapp setting? in case someone else looks at this

PR #1204, opened 2018-12-03
Rename courtOfferPage to courtOfferedInstalmentsPage

varma02:
### JIRA link
No Jira ticket - its just rename

### Change description
Rename courtOfferPage to courtOfferedInstalmentsPage


varma02:
@dharmendrak made a good point about url, i renamed that too just to make it more clearer

PR #1206, opened 2018-12-03
change Math.round to Lodash round in court determination files

varma02:
### JIRA link
No Jira

### Change description
For consistency, Math.round is refactored to use _.round in court determination files

### DIFF
--- a/src/main/features/claimant-response/routes/payment-option.ts
+++ b/src/main/features/claimant-response/routes/payment-option.ts
@@ -75,7 +75,6 @@ export class PaymentOptionPage extends AbstractPaymentOptionPage<DraftClaimantRe
         }
         return courtOfferedPaymentIntention
       }
-      return undefined
     }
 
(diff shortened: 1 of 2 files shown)


src/main/features/claimant-response/routes/payment-option.ts
```diff
@@ -75,7 +75,6 @@ export class PaymentOptionPage extends AbstractPaymentOptionPage<DraftClaimantRe
         }
         return courtOfferedPaymentIntention
       }
-      return undefined
```
sabahirfan: is this intentional ?

andrewwa-kainos: Yes, it's unreachable now. The prior code in this block is `if (something) { ... return } else { ... return } line 78`

varma02: yes, code never reaches there....i should have removed that as part of my previous PR. Noticed it now..

PR #1210, opened 2018-12-04
ROC-4831 - production bug fix for check and send error

rattur:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-4831

### Change description

Fix production check and send error when full admissions is disabled and user tries to go back to task list after hand off page

### Work checklist

- [ x] I have clicked through the running application to see if all changes I made actually work.

- [ ] Yes
- [x ] No


src/main/features/response/helpers/taskListBuilder.ts
```diff
@@ -58,13 +58,24 @@ export class TaskListBuilder {
...
-        Paths.responseTypePage.evaluateUri({ externalId: externalId }),
-        OweMoneyTask.isCompleted(draft)
+    if (!ClaimFeatureToggles.areAdmissionsEnabled(claim)
+      && (draft.isResponsePartiallyAdmitted() || draft.isResponseFullyAdmitted())) {
+      tasks.push(
+        new TaskListItem(
+          'Choose a response',
+          Paths.responseTypePage.evaluateUri({ externalId: externalId }),
+          false
+        )
```
andrewwa-kainos: Don't we want to be hiding the check and send instead?

rattur: No I'm afraid not after having a discussion with @robertparkinson

robertparkinson: @rattur cleaner solution would be when the users hit a paper handoff page from choose a response then delete the saved value for the choose a response page  from the draft object.

andrewwa-kainos: Or make check-and-send redirect to the hand-off page?

andrewwa-kainos: So then we could instead _not_ save the response option if we redirect to the hand-off page?

varma02: that may work, worth testing....It will be even smaller change if everything works

src/main/features/response/helpers/taskListBuilder.ts
```diff
@@ -57,13 +57,11 @@ export class TaskListBuilder {
...
     const externalId: string = claim.externalId
     const tasks: TaskListItem[] = []
-
     tasks.push(
       new TaskListItem(
         'Choose a response',
         Paths.responseTypePage.evaluateUri({ externalId: externalId }),
-        OweMoneyTask.isCompleted(draft)
-      )
+        OweMoneyTask.isCompleted(draft))
```
varma02: You didnt make any change in this file, PR will be cleaner if you just revert this file

src/main/features/response/views/check-and-send.njk
```diff
@@ -78,7 +78,7 @@
 
     {{ tableEnd() }}
 
-    {% if draft.isResponseFullyAdmitted() and not draft.fullAdmission.paymentIntention.paymentOption.isOfType(DefendantPaymentType.IMMEDIATELY) %}
+    {% if admissionsApplicable and draft.isResponseFullyAdmitted() and not draft.fullAdmission.paymentIntention.paymentOption.isOfType(DefendantPaymentType.IMMEDIATELY) %}
```
robertparkinson: remove this as users should not be able to get to the check and send page now

src/main/features/response/helpers/taskListBuilder.ts
```diff
@@ -33,6 +33,11 @@ export class TaskListBuilder {
     const tasks: TaskListItem[] = []
     const externalId: string = claim.externalId
 
+    if (!ClaimFeatureToggles.areAdmissionsEnabled(claim)
+      && (draft.isResponsePartiallyAdmitted() || draft.isResponseFullyAdmitted())) {
+      delete draft.response.type
+    }
+
```
varma02: Logic is fine but it feels it doesn't belong there...Please get another opinion @robertparkinson @andrewtaylorkainos @sabahirfan

rattur: It is only temporary until admissions is switched on. I have tried a couple areas to add this but this one seems appropriate at the moment

PR #1237, opened 2018-12-12
ROC-4893: Adding court decision on check-and-send page

singh-surjeet:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-4893

### Change description
Adding court decision on check-and-send page.

- [ ] Yes
- [X] No


src/main/features/claimant-response/views/check-and-send.njk
```diff
@@ -96,6 +96,66 @@
 
     {{ tableEnd() }}
 
+    {% set courtDetermination = draft.courtDetermination %}
+    {% set courtDecisionPaymentIntention = courtDetermination.courtDecision %}
+
+    {% if courtDetermination and courtDecisionPaymentIntention %}
```
robertparkinson: As discussed, please add a route test for this section that verifies that it shows when it is supposed too.

PR #1322, opened 2019-01-11
ROC-5037: convert aks manifests to helm

markry-kainos:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-5037

### Change description

Part of work for CCD integration. Although not specifically required for front-end - we should keep projects consistent.

- [ ] Yes
- [ ] No


charts/cmc-claim-store/requirements.yaml
```diff
@@ -0,0 +1,11 @@
+dependencies:
+  - name: postgresql
```
singh-surjeet: We happen to be maintaining postgres and java dependencies with different version than what we're using within claim-store?

markry-kainos: this is the helm chart version - i will look if we can pin postgres to same version as prod - i think it will be a case of defining the specific docker image in values.yaml 👍

PR #1360, opened 2019-01-24
ROC 5005 investigation when claim generation fails after successful payment 

tharacka:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-5005

### Change description

Applying app-insights to the pay screen to get further information to diagnose the issue

- [ ] Yes
- [x] No


src/main/features/claim/routes/pay.ts
```diff
@@ -141,8 +147,21 @@ export default express.Router()
       draft.document.claimant.payment = payment
       await new DraftService().save(draft, user.bearerToken)
 
+      trackCustomEvent(`Pre payment for externalId: ${externalId}`,{
```
ellie-harrison: I dont think there's any value in tracking this event, we only really need to know when this fails.

src/main/features/claim/routes/pay.ts
```diff
@@ -165,6 +184,12 @@ export default express.Router()
 
       await new DraftService().save(draft, user.bearerToken)
 
+      trackCustomEvent(`Post payment for externalId: ${externalId}`,{
```
ellie-harrison: Same here, I think there's only value in tracking the failure

src/main/app/logging/customEventTracker.ts
```diff
@@ -1,10 +1,14 @@
...
+export function trackCustomEvent (eventName: string, trackingProperties: {}) {
+  try {
+    if (appInsights.defaultClient) {
+      appInsights.defaultClient.trackEvent({
+        name: eventName,
+        properties: trackingProperties
+      })
+    }
+  } catch (err) {
+    return
```
ellie-harrison: Should we be doing something with this error than just returning?

src/main/features/claim/routes/pay.ts
tharacka: The reason why I originally put them in is to ensure that the microservices being used in pay.ts were behaving. Also, I also wanted to make sure that the RC emails being generated reflected the correct status. However, for the time being I see your point and have removed them

src/main/app/logging/customEventTracker.ts
tharacka: As above, have been removed

src/main/features/claim/routes/pay.ts
tharacka: As above

PR #1367, opened 2019-01-29
ROC-4863 You can only use mediation if

robertparkinson:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-4864

### Change description

New page in the mediation flow

### Work checklist

- [ ] Yes
- [x] No


robertparkinson:
Both the response and claimant-response use mediation and there is now going to be many more pages. I thought it would make more sense to separate out the mediation journey similar to how we have done the CCJ journey as it can be gotten to from multiple ways.

KiranVarma02:
I have approved but as discussed please get another opinion from @sabahirfan or @dharmendrak

src/main/features/mediation/routes/mediation-agreement.ts
```diff
@@ -0,0 +1,52 @@
...
+        if (!claim.isResponseSubmitted()) {
+          res.redirect(ResponsePaths.taskListPage.evaluateUri({ externalId: claim.externalId }))
+        } else {
+          res.redirect(ClaimantResponsePaths.taskListPage.evaluateUri({ externalId: claim.externalId }))
+        }
+      } else {
+        draft.document.youCanOnlyUseMediation = new FreeMediation(FreeMediationOption.YES)
+
+        await new DraftService().save(draft, user.bearerToken)
+        // TODO update to point to next page when it is ready
```
sabahirfan: do we have story for that in flight?

src/main/features/mediation/routes/mediation-agreement.ts
```diff
@@ -0,0 +1,52 @@
...
+    renderView(res)
+  })
+  .post(
+    Paths.mediationAgreementPage.uri,
+    ErrorHandling.apply(async (req: express.Request, res: express.Response) => {
+      const claim: Claim = res.locals.claim
+      const draft: Draft<MediationDraft> = res.locals.mediationDraft
+      const user: User = res.locals.user
+
+      if (req.body.reject) {
```
sabahirfan: The current logic may assume it is accepted if got post from any other place (other than `mediation-agreement.njk`.
How about deciding if we have `req.body.accept` rather than `reject`? It will be more safe.

src/main/features/mediation/routes/mediation-agreement.ts
robertparkinson: yes. it's in progress at the moment.

PR #1375, opened 2019-01-31
ROC-5110 Claimant accept states paid less than amount claim, answers not saved in DB

ellie-harrison:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-5110

### Change description
Pass through answers to states paid questions to claim store on submission

- [x] Yes
- [ ] No


src/main/app/claims/converters/claimantResponseConverter.ts
```diff
@@ -106,4 +111,25 @@ export class ClaimantResponseConverter {
...
+
+  private static addStatesPaidOptions (draftClaimantResponse: DraftClaimantResponse,
+                                       claimantResponse: ClaimantResponse) {
+    if (draftClaimantResponse.partPaymentReceived) {
+      claimantResponse.paymentReceived =
+        this.convertYesNoString(draftClaimantResponse.partPaymentReceived.received.option)
+    }
+
+    if (draftClaimantResponse.accepted) {
+      claimantResponse.settleForAmount = this.convertYesNoString(draftClaimantResponse.accepted.accepted.option)
```
sabahirfan: I think we don't need `convertYesNoString` util method, How about?
```
draftClaimantResponse.accepted.accepted.option as YesNoOption // ( this is from core/yesNoOption)
```

PR #1383, opened 2019-02-05
ROC-5084 Error thrown when clicked on claim from claimant dashboard

ellie-harrison:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-5084

### Change description
Fixed issue causing an error when viewing the dashboard after a claimant accepts or rejects a states paid response

- [ ] Yes
- [x] No


src/main/app/claims/models/claim.ts
```diff
@@ -302,7 +307,9 @@ export class Claim {
       return true
     }
 
-    return (((this.response && (this.response as FullAdmissionResponse).paymentIntention.paymentOption !== PaymentOption.IMMEDIATELY && !this.isSettlementReachedThroughAdmission()
+    return (((this.response && (this.response as FullAdmissionResponse).paymentIntention &&
```
markry-kainos: just feels messy :))) not sure what to suggest though? maybe breaking out into const booleans or refactor into helper methods somewhere? its fine though...so maybe we do as refactor task another day

sabahirfan: Helper method would be preferred here.

PR #1452, opened 2019-02-26
Integrating the key vault flex volume for secrets.

KevinJCross:
### JIRA link
https://tools.hmcts.net/jira/browse/RPE-951

### Change description
This is part of the work from the RPE team to move towards using secrets directly from AKS using a flex volume to avoid Environment variables (Read from the file system directly). These changes will be needed before we can consider going to live and also allow the access of multiple keyvaults within the secrets object.

This required changes:
- update of the helm chart
- moving all secrets to the "secrets" configuration object
- updating to make sure the secrets are defaulted to the environment variables if available.
- calling the `property-volume.addTo`

- [ ] Yes
- [x] No


Jenkinsfile_CNP
```diff
@@ -6,16 +6,10 @@ String product = "cmc"
 String component = "citizen-frontend"
 
 List<LinkedHashMap<String, Object>> secrets = [
+  //for smoke tests
```
timja: not sure if this is needed but its smoke / functional tests

src/integration-test/helpers/clients/idamClient.ts
```diff
@@ -9,6 +9,7 @@ const defaultPassword = 'Password12'
 const oauth2 = {
   client_id: 'cmc_citizen',
   redirect_uri: `${process.env.CITIZEN_APP_URL}/receiver`,
+  // integration test ?? .. secrets.cmc.citizen-oauth-client-secret
```
timja: remove now?

Jenkinsfile_CNP
KevinJCross: I believe it is for the docker images that are run up for smoke/functional.

PR #1465, opened 2019-03-05
Fix aseName to not use remote state

markry-kainos:
### Change description

core-compute-aat remote state nuked by accident. This PR to fix sandbox deployments but should we fix permanently?

- [ ] Yes
- [x] No


infrastructure/main.tf
```diff
@@ -13,7 +13,7 @@ provider "vault" {
 }
 
 locals {
-  aseName = "${data.terraform_remote_state.core_apps_compute.ase_name[0]}"
+  aseName = "core-compute-${var.env}"
```
timja: same as: ccd-user-profile-api-POSTGRES-PASS

markry-kainos: as a secret? 🤔 not sure...does it matter 😺

PR #1481, opened 2019-03-08
ROC-5258 Refactor claim issue dashboard

simad:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-5258

### Change description
- Introducing a new way of discerning the dashboard status from a claim. It will still be picked based on the state of a claim (as it is now), but will be calculated using a 'flowchart'-like data structure instead of complex if-elses. This should make everything more testable and also it should be easier to follow what's going on.
- Added some additional guards to make sure we default to the old way if the new rule engine can't find a dashboard status / the rule engine itself is wrong. Those will be removed once we have migrated all the statuses and we are confident the rule engine works.
- Added a tslint rule to stop committing `.only` in tests (guilty!)

In more detail: 
- Each element of this flowchart has a `description` (for documentation purposes), an optional `dashboard` (the string name of the `njk` template containing the status), a `condition` (a function which is true when the claim is in that status) and `next` states.
 Basically, the code will go through the flowchart, given a certain claim. It will evaluate the condition of each state, and if it's a leaf (there are no next states, or all of next states are `false`) then it will return the dashboard, otherwise it will go to the next valid state and so on.
 - Only one of the flowchart paths should be valid at any given time - if there are two possible paths, it means that either the conditions in the flowcharts are wrong, or the claim is in a weird state
 - Each dashboard status should go to a separate nunjucks template. This will make things easier to reason about as the logic is 100% in typescript, rather than in two places. To not take this to the extreme, tho, there shouldn't be a problem having super simple conditions in nunjucks - when the logic becomes too cumbersome, it should definitely be moved in the "rule engine"

### Work checklist

- [ ] Yes
- [x] No

### DIFF
--- a/src/main/features/dashboard/helpers/claimStatusFlow.ts
+++ b/src/main/features/dashboard/helpers/claimStatusFlow.ts
@@ -6,7 +6,7 @@ const logger = Logger.getLogger('ClaimStatusFlow')
 export class ClaimStatusFlow {
 
   static readonly flow: ClaimStatusNode = {
-    description: 'Claim Is Draft',
+    description: 'Claim Exists',
     isValidFor: () => true,
     next: [
       {
@@ -20,9 +20,13 @@ export class ClaimStatusFlow {
   static decide (flow: ClaimStatusNode, claim: Claim): string {
     if (flow.isValidFor(claim)) {
       const nextPossibleConditions = (flow.next || []).filter(state => state.isValidFor(claim))
-      if (nextPossibleConditions.length > 1) throw new Error(`Two possible paths are valid for a claim, check the flow's logic`)
+      if (nextPossibleConditions.length > 1) {
+        throw new Error(`Two possible paths are valid for a claim, check the flow's logic`)
+      }
       if (nextPossibleConditions.length === 0) {
-        if (!flow.dashboard) throw new Error(`Trying to render an intermediate state with no dashboard, check the flow's logic`)
+        if (!flow.dashboard) {
+          throw new Error(`Trying to render an intermediate state with no dashboard, check the flow's logic`)
+        }
         return flow.dashboard
       }
       return this.decide(nextPossibleConditions[0], claim)
--- a/src/main/features/dashboard/helpers/claimStatusFlow.ts
+++ b/src/main/features/dashboard/helpers/claimStatusFlow.ts
@@ -1,0 +1,49 @@
+import { Claim } from 'claims/models/claim'
+import { Logger } from '@hmcts/nodejs-logging'
+
+const logger = Logger.getLogger('ClaimStatusFlow')
+
+export class ClaimStatusFlow {
+
+  static readonly flow: ClaimStatusNode = {
+    description: 'Claim Is Draft',
+    isValidFor: () => true,
+    next: [
+      {
+        description: 'Claim Issued',
+        isValidFor: (claim) => !claim.response,
+        dashboard: 'claim_issued',
+        next: []
+      }
+    ]
+  }
+  static decide (flow: ClaimStatusNode, claim: Claim): string {
+    if (flow.isValidFor(claim)) {
+      const nextPossibleConditions = (flow.next || []).filter(state => state.isValidFor(claim))
+      if (nextPossibleConditions.length > 1) throw new Error(`Two possible paths are valid for a claim, check the flow's logic`)
+      if (nextPossibleConditions.length === 0) {
+        if (!flow.dashboard) throw new Error(`Trying to render an intermediate state with no dashboard, check the flow's logic`)
+        return flow.dashboard
+      }
+      return this.decide(nextPossibleConditions[0], claim)
+    }
+  }
+
+  // the try/catch should be removed once we don't need backward compatibility with 'old' dashboards - it's here to make sure we render the
+  // old status in case there are problems with the new one (which should be addressed)
+  public static dashboardFor (claim: Claim): string {
+    try {
+      return this.decide(ClaimStatusFlow.flow, claim)
+    } catch (err) {
+      logger.error(err)
+      return ''
+    }
+  }
+}
+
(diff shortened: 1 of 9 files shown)


src/main/features/dashboard/helpers/claimStatusFlow.ts
```diff
@@ -0,0 +1,49 @@
+import { Claim } from 'claims/models/claim'
+import { Logger } from '@hmcts/nodejs-logging'
+
+const logger = Logger.getLogger('ClaimStatusFlow')
+
+export class ClaimStatusFlow {
+
+  static readonly flow: ClaimStatusNode = {
+    description: 'Claim Is Draft',
```
simad: @andrewwa-kainos made me aware that this is not really true (as drafts are handled in a different way compared to claims), so this description is not correct. It should be more like `Claim exists`, it's just a root state which kicks off the whole flow.  I'll change that when I have to address the review comments which will follow :D

src/main/features/dashboard/index.ts
```diff
@@ -20,6 +27,17 @@ function requestHandler (): express.RequestHandler {
...
 }
 
+function render (claim: Claim, type: string): string {
+  const dashboardName = ClaimStatusFlow.dashboardFor(claim)
+  try {
+    const template = nunjucks.render(path.join(__dirname, './views', 'status', type, dashboardName + '.njk').toString(), { claim: claim })
+    return app.settings.nunjucksEnv.filters['safe'](template)
+  } catch (err) {
+    logger.error(`view not found for status: ${dashboardName}`)
+    return ''
```
KiranVarma02: wondering if there is any "something has gone wrong" kind of status page will be useful here!

@robertparkinson do we have any invalid status page??

robertparkinson: at the moment production just shows a blank status if we've missed one. same for the case details view. Probably better to show some sort of error to the user

simad: yes, that is an option - right now returning `' '` will default to the "old way" of doing things - but if something is not covered in the old way, then it's going to be blank

src/test/features/dashboard/helpers/claimStatusFlow.ts
```diff
@@ -0,0 +1,135 @@
...
+      const flow: ClaimStatusNode = {
+        description: 'this is always true',
+        isValidFor: () => true,
+        dashboard: 'first',
+        next: []
+      }
+      const claim: Claim = new Claim().deserialize(sampleClaimObj)
+      expect(ClaimStatusFlow.decide(flow, claim)).to.equal(`first`)
+    })
+  })
```
KiranVarma02: Please add line spacings between tests just for readability (We follow this in other tests). Everything else looks good..

src/main/features/dashboard/helpers/claimStatusFlow.ts
```diff
@@ -0,0 +1,49 @@
...
+        isValidFor: (claim) => !claim.response,
+        dashboard: 'claim_issued',
+        next: []
+      }
+    ]
+  }
+  static decide (flow: ClaimStatusNode, claim: Claim): string {
+    if (flow.isValidFor(claim)) {
+      const nextPossibleConditions = (flow.next || []).filter(state => state.isValidFor(claim))
+      if (nextPossibleConditions.length > 1) throw new Error(`Two possible paths are valid for a claim, check the flow's logic`)
```
andrewwa-kainos: @robertparkinson do we have a policy for TS as well as Java to have all code in `{` blocks `}`?

src/main/features/dashboard/index.ts
```diff
@@ -9,6 +10,12 @@ import { DraftClaim } from 'drafts/models/draftClaim'
 import { OAuthHelper } from 'idam/oAuthHelper'
 import { PaymentSchedule } from 'claims/models/response/core/paymentSchedule'
 import { Paths } from 'dashboard/paths'
+import { Claim } from 'claims/models/claim'
+import { ClaimStatusFlow } from 'dashboard/helpers/claimStatusFlow'
+import { Logger } from '@hmcts/nodejs-logging'
+import { app } from '../../app'
```
andrewwa-kainos: Please avoid relative paths if possible

simad: dammit, IntelliJ auto imports :) will refactor 👍

src/test/features/dashboard/helpers/claimStatusFlow.ts
simad: 👍

src/main/features/dashboard/helpers/claimStatusFlow.ts
robertparkinson: Yeah we do keep everything inside { }

src/test/http-mocks/claim-store.ts
```diff
@@ -401,10 +401,10 @@ export function resolveRetrieveClaimByExternalIdTo404HttpCode (reason: string =
     .reply(HttpStatus.NOT_FOUND, reason)
 }
 
-export function resolveRetrieveByClaimantId (claimOverride?: object) {
+export function resolveRetrieveByClaimantId (claim: object = sampleClaimObj, claimOverride?: object) {
```
KiranVarma02: are you doing this to be able to use it for other claim objects if needed in future?

simad: I did this so that I didn't have to modify all of the (hundreds?) of route tests currently using that method with the old signature

simad: But yes, it thought it does not make so much sense to keep duplicating methods when the only difference is a single object, it's better to just pass it in

src/main/features/dashboard/helpers/claimStatusFlow.ts
simad: should be fixed 👍

PR #1482, opened 2019-03-08
ROC-5241 Defendant Hearing - Choose a hearing location - postcode lookup

ellie-harrison:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-5241

### Change description
Add client for the court and tribunal finder API

- [ ] Yes
- [x] No


src/main/app/court-finder/areaOfLaw.ts
```diff
@@ -0,0 +1,46 @@
...
+      AreaOfLaw.HOUSING_POSSESSION,
+      AreaOfLaw.IMMIGRATION,
+      AreaOfLaw.MONEY_CLAIMS,
+      AreaOfLaw.PROBATE,
+      AreaOfLaw.SOCIAL_SECURITY,
+      AreaOfLaw.TAX,
+      AreaOfLaw.ALL
+    ]
+  }
+}
```
KiranVarma02: Just want to understand, do we ever use other AreaOfLaw in our case?

src/main/app/court-finder/courtFinderClient.ts
```diff
@@ -0,0 +1,83 @@
...
+            court.dx_number,
+            court.distance,
+            new Address(
+              court.address.address_lines,
+              court.address.postcode,
+              court.address.town,
+              court.address.type),
+            court.areas_of_law.map((aol: any) => {
+              return new AreaOfLaw(aol.name, aol.external_link, aol.external_link_desc)
+            }))
```
KiranVarma02: Just wondering, do we need to get everything since we only need Address? What do you think?

KiranVarma02: we could simply code and test if we only get Address...

ellie-harrison: It would be ever so slightly simpler if we only got the address, but theres no maintanence overhead for us to pull everything currently. It was originally build with it being an NPM module in mind for resusability, personally I'd prefer to leave it this way so it can be factored out for reuse with minimal work required if other people want to use it in the future. But I'm open to whatever.

PR #1485, opened 2019-03-11
ROC-5002 filter tests

andrewwa-kainos:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-5002

### Change description
Add unit tests for nunjucks filters

### Work checklist

- [ ] Yes
- [x] No


src/test/features/claimant-response/filters/frequency-view-filter.ts
```diff
@@ -0,0 +1,22 @@
...
+        })
+      })
+
+    it('should throw an error for null', () => {
+      expect(() => FrequencyViewFilter.render(null)).to.throw(TypeError)
+    })
+  })
+
+  context('render payment frequency', () => {
+    // TODO: https://tools.hmcts.net/jira/browse/ROC-5325
```
robertparkinson: is there a way to write the so it will pass now but fail when the bug is fixed?

andrewwa-kainos: As discussed on slack: I will update 5325 to remind whoever works it to add these tests

src/test/features/claimant-response/filters/bank-account-type-view-filter.ts
```diff
@@ -0,0 +1,20 @@
+import { BankAccountType } from 'features/response/form/models/statement-of-means/bankAccountType'
+import { expect } from 'chai'
+import { BankAccountTypeViewFilter } from 'claimant-response/filters/bank-account-type-view-filter'
+
+describe('Bank account type view filter', () => {
+  BankAccountType.all()
+    .forEach(type => {
```
simad: this is not strictly related to this PR, but i noticed that we are not actually testing the values / displayValue of the enums. Meaning, 
`static readonly CURRENT_ACCOUNT =  new BankAccountType('CURRENT_ACCOUNT', 'Current account')`
if I change the `displayName` of `Current account` to `I need coffee`, all of the tests will still pass, which might be not ideal. 
It's something we might consider adding (again, does not have to be part of this PR :) )

simad: I have just realised we do a more thorough testing in this file https://github.com/hmcts/cmc-citizen-frontend/pull/1485/files#diff-cf49c310d1461059dc7cd595cfe84e01 
I'd suggest - it's ok to merge the PR as it is, but then we'd probably need to add unit tests for the enum themselves, or we can 'explode' this `forEach` testing each row in a test like you did in the above file. What do you think, @andrewwa-kainos ?

andrewwa-kainos: I would argue the filter's job is to take _whatever_ the enum has. Changing an enum value to `I need coffee`, however true that statement might be, should impact the UI and be picked up by route tests when that's all in place. What do you think, @simad ?

robertparkinson: I would agree with how andy has done it but silvia is right that we should have some unit tests testing the content of the enums so they are at least verifies somewhere. we don't currently test all the content on each of the pages in the route tests.

andrewwa-kainos: Resolving on that basis then.

simad: @andrewwa-kainos I agree with you when you say that the filter's job should take any value, as long as it matches. That's why I suggested keeping the current approach and adding unit tests (rather than route tests) to the enums themselves. Route tests are fairly more complicated than simple unit tests and I feel the "lower" level testing we do, the fastest and more thorough the feedback :)

PR #1514, opened 2019-03-18
Update jenkinsfiles with new secrets format

markry-kainos:
### Change description

Sandbox deployments are broken because secrets not updated in-line with Jenkinsfile_CNP. Made necessary changes to nightly and paramaterised files.

- [ ] Yes
- [x] No


markry-kainos:
@sabahirfan think we will need to do something for legal too

PR #1539, opened 2019-03-27
ROC-5160 Split Full Name for Their Details (individuals and sole traders)

simad:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-5160

### Change description
Splitting the full name box to title / first name / last name for individuals and sole traders, their details. Whenever first name and last name are provided, the frontend will use those, concatenating them with the title if present, otherwise it will use the full name.
This should make sure old claims which only have the name will still be valid. 

New claims which are currently in draft and have a full name, won't be able to be submitted and the user will have to re-insert their details in the right format.

### Work checklist

- [ ] Yes
- [x] No


src/main/app/forms/models/soleTraderDetails.ts
```diff
@@ -1,16 +1,35 @@
...
+    return `${input} must be no longer than $constraint1 characters`
+  }
 }
 
 export class SoleTraderDetails extends PartyDetails {
 
   @MaxLength(35, { message: ValidationErrors.ORGANISATION_NAME_TOO_LONG, groups: ['claimant', 'defendant'] })
   businessName?: string
 
+  title?: string
```
andrewwa-kainos: No validation on title length?

simad: Good catch, added it :) thanks!

src/integration-test/data/test-data.ts
```diff
@@ -146,11 +146,13 @@ export function createDefendant (type: PartyType, hasEmailAddress: boolean = fal
 
   switch (type) {
     case PartyType.INDIVIDUAL:
-      defendant.name = 'Rose Smith'
+      defendant.firstName = 'Rose'
```
sabahirfan: Would you want to add title here as well?

src/test/app/forms/models/individualDetails.ts
```diff
@@ -6,13 +6,15 @@ import { PartyType } from 'common/partyType'
...
 import { ValidationErrors as CorrespondenceAddressValidationErrors } from 'forms/models/correspondenceAddress'
 import { ValidationError, Validator } from '@hmcts/class-validator'
-import { expectValidationError } from 'test/app/forms/models/validationUtils'
+import { expectValidationError, expectValidationErrorNotPresent } from 'test/app/forms/models/validationUtils'
 import { LocalDate } from 'forms/models/localDate'
+import { ValidationErrors as SplitNamedPartyDetailsValidationErrors } from 'forms/models/splitNamedPartyDetails'
+
 const validAddress = new Address('line1', 'line2', 'line3', 'city', 'bb127nq')
 
 const aVeryLongString = (): string => {
```
sabahirfan: Would be nice to replace this with `validationUtils.generateString`

src/integration-test/data/test-data.ts
simad: I was thinking it would be ok not to test for title being optional, since I test for its presence in almost all of the unit tests. Lemme think

src/test/app/forms/models/individualDetails.ts
simad: 👍 didn't add that but yeah, if we already have something in place it makes sense to use it

src/integration-test/data/test-data.ts
simad: added it to individual but not sole trader, so we still have a journey checking that everything is fine if title is not provided

simad: (ideally I'd add two different journeys for both, but I think this is an ok compromise :D)

sabahirfan: That's fine.

src/test/app/forms/models/individualDetails.ts
sabahirfan: Thank you

src/main/app/pdf/mappers/theirDetailsMapper.ts
```diff
@@ -43,6 +44,28 @@ export class TheirDetailsMapper {
     }
   }
 
+  private static splitNames (partyDetails: TheirDetails): object {
+    if (partyDetails && partyDetails.type) {
+      switch (partyDetails.type) {
```
robertparkinson: is there any benefit to having this switch? you're returning the base partyDetails properties so you could just cast as PartyDetails or just as Individual and return the same title, firstName, lastName. there's nothing specific to Individual or SoleTrader that is being sent

src/test/app/claims/claimModelConverter.ts
```diff
@@ -65,4 +72,13 @@ describe('ClaimModelConverter', () => {
     const converted: ClaimData = ClaimModelConverter.convert(claimDraft)
     expect(converted.interest.interestDate).to.be.undefined
   })
+
+  it('should not contain title if blank', () => {
+    const claimDraft = prepareClaimDraft(theirIndividualDetails, individualDefendant)
+    // @ts-ignore
```
robertparkinson: why the ts-ignores? can we not write it in a way to pass the linter?

src/main/app/pdf/mappers/theirDetailsMapper.ts
simad: you're right. we have a similar switch in all of the methods of this class (even when there's only one clause) and I was being foolishly consistent :)

src/test/app/claims/claimModelConverter.ts
simad: Because being it a test my laziness took over and I let intellij put the ignores, rather than spending literally 3 seconds trying to appease the linter properly. Sorry :D

PR #1557, opened 2019-04-02
ROC-5433-Remove-unnecessary-fetch-call-before-save

karthikkannankainos:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-5433

### Change description

Remove fetch case by external id unnecessary before save.


src/main/features/claim/routes/pay.ts
```diff
@@ -102,8 +74,30 @@ async function successHandler (res, next) {
...
     }
     await claimStoreClient.saveClaim(draft, user, features)
-  }
 
+  } catch (err) {
+    if (err.statusCode === HttpStatus.INTERNAL_SERVER_ERROR || err.statusCode === HttpStatus.SERVICE_UNAVAILABLE) {
+      logError(
+        user.id,
+        draft.document.claimant.payment,
+        `Payment processed successfully but there is problem saving claim from claim store externalId: ${externalId}.`
```
andrewwa-kainos: Can we fix the grammar please?
`... there was a problem saving claim '${externalId}' to the claim store`

src/main/features/claim/routes/pay.ts
```diff
@@ -102,8 +74,30 @@ async function successHandler (res, next) {
...
+          payment: draft.document.claimant.payment,
+          error: err
+        })
+      next(err)
+      return
+    } else if (err.statusCode === HttpStatus.CONFLICT) {
+      logError(
+        user.id,
+        draft.document.claimant.payment,
+        `Payment processed successfully and claim already exist with same externalId: ${externalId}.`
```
andrewwa-kainos: Can we fix the grammar please?
`... and claim '${externalId}' already exists.`

src/main/features/claim/routes/pay.ts
```diff
@@ -58,35 +58,7 @@ async function successHandler (res, next) {
   const user: User = res.locals.user
   const externalId: string = draft.document.externalId
 
-  let claimIsAlreadyFullyPersisted: boolean
-
   try {
```
KiranVarma02: Could i please ask, what is the reason earlier to check if claim is already fully persisted? and why we think its not necessary now? This is for my understanding...

src/main/features/claim/routes/pay.ts
karthikkannankainos: done

src/main/features/claim/routes/pay.ts
karthikkannankainos: Done @andrewwa-kainos .

src/main/features/claim/routes/pay.ts
karthikkannankainos: @KiranVarma02 , we felt that making call is unnecessary as we are having transaction bound call.

KiranVarma02: ok, thanks for explaining

PR #1567, opened 2019-04-09
Use IDAM PR Helm chart for whitelisting urls

markry-kainos:
### Change description

Testing a cleaner way to manage SIDAM whitelisting for PR URLs. This will add on install and remove on delete (merge).

- [ ] Yes
- [x] No


markry-kainos:
@timja sorry this took another day! but it works tested
1. manual login via postman - fail as expected
2. ^^^ again after install to aks - work as expected (smoke/functional pass too)
3. and again after manual `helm del --purge` - fail as expected

So good to go!

PR #1592, opened 2019-04-24
RPE-1043 Launchdarkly integration

klawani:
### JIRA link

https://tools.hmcts.net/jira/browse/RPE-1043

### Change description

This task is to integrate LaunchDarkly product as a replacement for the current FeatureToggling API currently being used.

The current implementation is at early stage and changes will be made as we progress.

### Work checklist

- [x ] Yes
- [ ] No


config/default.json
```diff
@@ -108,5 +108,13 @@
       "encryptionKey": "not-a-real-32-bit-encryption-key",
       "AppInsightsInstrumentationKey":"APPINSIGHTS_INSTRUMENTATIONKEY"
     }
+  },
+  "launchDarkly": {
+    "featureFlag-keys": {
```
timja: missing of camel case and hiphen looks weird
maybe:
```suggestion
    "keys": {
```

not 100% convinced this needs to be in config, they need to be static everywhere a constants file maybe be better

config/default.json
```diff
@@ -108,5 +108,13 @@
       "encryptionKey": "not-a-real-32-bit-encryption-key",
       "AppInsightsInstrumentationKey":"APPINSIGHTS_INSTRUMENTATIONKEY"
     }
+  },
+  "launchDarkly": {
```
timja:
```suggestion
  "featureFlags": {
```

config/default.json
```diff
@@ -108,5 +108,13 @@
...
+    "featureFlag-keys": {
+      "admissions": "cmc_admissions"
+    },
+    "sdk-keys": {
+      "production": "sdk-a9efb423-1040-4dfd-b0b7-92d4e1898f7a"
```
timja: shouldn't commit this, can you revoke and rotate and bring in with an environment variable?

config/default.json
```diff
@@ -108,5 +108,13 @@
       "encryptionKey": "not-a-real-32-bit-encryption-key",
       "AppInsightsInstrumentationKey":"APPINSIGHTS_INSTRUMENTATIONKEY"
     }
+  },
+  "launchDarkly": {
+    "featureFlag-keys": {
+      "admissions": "cmc_admissions"
+    },
+    "sdk-keys": {
```
timja: you don't need different keys per env, just one and use an environment variable to bring it in, see 
https://github.com/hmcts/cmc-citizen-frontend/blob/master/config/custom-environment-variables.json

src/main/common/clients/launchDarklyClient.ts
```diff
@@ -0,0 +1,25 @@
...
+const sdkKeys: string = config.get<string>('launchDarkly.sdk-keys.production')
+
+export class LaunchDarklyClient {
+  static client: ld.LDClient
+  static initClient () {
+    this.client = ld.init(sdkKeys)
+  }
+
+  callFeatureFlag (user: User, featureKey: string, f?: () => void) {
+    let ldUser: ld.LDUser = {
```
timja:
```suggestion
    const ldUser: ld.LDUser = {
```

src/main/common/clients/launchDarklyClient.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const sdkKeys: string = config.get<string>('launchDarkly.sdk-keys.production')
+
+export class LaunchDarklyClient {
+  static client: ld.LDClient
+  static initClient () {
+    this.client = ld.init(sdkKeys)
+  }
+
+  callFeatureFlag (user: User, featureKey: string, f?: () => void) {
```
timja: add return type
```suggestion
  callFeatureFlag (user: User, featureKey: string, f?: () => void): void {
```

src/main/common/clients/launchDarklyClient.ts
```diff
@@ -0,0 +1,25 @@
+import * as config from 'config'
+import { User } from 'idam/user'
+import * as ld from 'ldclient-node'
+
+const sdkKeys: string = config.get<string>('launchDarkly.sdk-keys.production')
+
+export class LaunchDarklyClient {
+  static client: ld.LDClient
+  static initClient () {
```
timja: think about whether an instance using a constructor would make more sense, not sure

src/main/features/claim/routes/pay.ts
```diff
@@ -93,10 +95,10 @@ async function successHandler (res, next) {
       logger.error(`missing consent not given role for user, User Id : ${user.id}`)
     }
 
-    let features: string
-    if (await featureTogglesClient.isFeatureToggleEnabled(user, roles, 'cmc_admissions')) {
+    let features: string = ''
+    launchDarklyClient.callFeatureFlag(user, config.get<string>('launchDarkly.featureFlag-keys.admissions'), function () {
```
timja: this config.get is too long either extract to a variable or make it a constant

also do you need an anonymous function or can you use an arrow function?
```suggestion
    launchDarklyClient.callFeatureFlag(user, config.get<string>('launchDarkly.featureFlag-keys.admissions'), () => {
```

timja:
could you create an `ithc` branch and repoint the base branch to that as this isn't intended for master

src/main/features/claim/routes/pay.ts
```diff
@@ -93,10 +95,10 @@ async function successHandler (res, next) {
       logger.error(`missing consent not given role for user, User Id : ${user.id}`)
     }
 
-    let features: string
-    if (await featureTogglesClient.isFeatureToggleEnabled(user, roles, 'cmc_admissions')) {
+    let features: string = ''
+    launchDarklyClient.callFeatureFlag(user, config.get<string>('launchDarkly.featureFlag-keys.admissions'), () => {
```
robertparkinson: Can we change it so that we just update the `isFeatureToggleEnabled` method to call the new LaunchDarklyClient so we only need to change one place in our code?

timja: the calling method is different I think, it takes a closure rather than responding with a boolean, so probably requires refactoring either way

klawani: @timja is right ... a refactoring is needed either way.

robertparkinson: I've no doubt that it will require refactoring. I'm asking for that to be done inside the isFeatureToggleEnabled so that it will continue to return a boolean :)

config/custom-environment-variables.json
```diff
@@ -98,5 +98,12 @@
       "encryptionKey": "COOKIE_ENCRYPTION_KEY",
       "AppInsightsInstrumentationKey":"APPINSIGHTS_INSTRUMENTATIONKEY"
     }
+  },
+
+  "launchDarkly": {
+    "sdk": "LD_SDK_KEYS",
+    "featureFlags": {
+       "admissions": "LD_FEATURE_ADMISSIONS"
```
timja: no need for this to be configurable per env, debatable on whether it should be in config at all

package.json
```diff
@@ -77,6 +77,7 @@
     "i18next-sprintf-postprocessor": "^0.2.2",
     "jquery": "3.4.0",
     "js-base64": "^2.5.1",
+    "ldclient-node": "^5.7.4",
```
timja: this should have been added with `yarn add ldclient-node` I assume you manually added it as I can't see any yarn.lock file changes in this PR?

src/main/common/clients/launchDarklyClient.ts
klawani: I have now implemented this as a Singleton as per recommendation from LaunchDarkly documentation (https://docs.launchdarkly.com/docs/node-sdk-reference)

config/default.json
klawani: Move config to ENV variable and ADD to terraform

config/default.json
klawani: This should come now from KeyVault

src/main/features/claim/routes/pay.ts
```diff
@@ -93,10 +95,11 @@ async function successHandler (res, next) {
       logger.error(`missing consent not given role for user, User Id : ${user.id}`)
     }
 
-    let features: string
-    if (await featureTogglesClient.isFeatureToggleEnabled(user, roles, 'cmc_admissions')) {
+    let features: string = ''
+    const featureKeys: string = config.get<string>('launchDarkly.featureFlags.admissions')
+    launchDarklyClient.callFeatureFlag(user, featureKeys , () => {
```
timja: is this synchronous or async?
i.e. will it return before the next code block is executed?

config/custom-environment-variables.json
klawani: that is true

config/default.json
```diff
@@ -108,5 +108,9 @@
       "encryptionKey": "not-a-real-32-bit-encryption-key",
       "AppInsightsInstrumentationKey":"APPINSIGHTS_INSTRUMENTATIONKEY"
     }
+  },
+
+  "launchDarkly": {
+    "sdk": "sdk-AAAAA"
```
timja: this needs to go in the secrets block and the key name in vault needs to match this 
https://github.com/hmcts/cmc-citizen-frontend/blob/2a2987ca5509385505f28a601b4a7e6b9257cfd7/config/default.json#L100

package.json
klawani: done

src/main/features/claim/routes/pay.ts
klawani: make function to be Async now.

src/main/common/clients/launchDarklyClient.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  static initClient () {
+    if (!LaunchDarklyClient.client) {
+      LaunchDarklyClient.client = ld.init(sdkKeys)
+    }
+  }
+
+  async callFeatureFlag (user: User, featureKey: string, f?: () => void) {
+    const ldUser: ld.LDUser = {
+      key: user.id
```
timja: I think you need to pass the user roles here as well

src/main/features/claim/routes/pay.ts
```diff
@@ -93,10 +95,10 @@ async function successHandler (res, next) {
       logger.error(`missing consent not given role for user, User Id : ${user.id}`)
     }
 
-    let features: string
-    if (await featureTogglesClient.isFeatureToggleEnabled(user, roles, 'cmc_admissions')) {
+    let features: string = ''
+    await launchDarklyClient.callFeatureFlag(user, 'cmc_admissions' , () => {
```
timja: this isn't quite the same as before, the toggle is configured to only sometimes say yes and only if the user has the role `cmc-new-features-consent-given`

src/main/features/claim/routes/pay.ts
timja: not sure if that is worth doing, you'll likely miss out on some of the SDK features if you wrap it in this way,

config/default.json
klawani: done

src/main/common/clients/launchDarklyClient.ts
klawani: I have added it to LDUser as custom attributes

src/main/features/claim/routes/pay.ts
klawani: I have added it to LDUser as custom attributes

markry-kainos: yes!!!!! sorry i forgot about that @klawani 🙈

package.json
```diff
@@ -94,8 +95,8 @@
     "to-boolean": "^1.0.0",
     "ts-node": "^8.1.0",
     "tsconfig-paths": "^3.8.0",
-    "typescript": "^3.4.5",
     "tslint-microsoft-contrib": "^6.1.1",
+    "typescript": "^2.9.2",
```
timja: why was this changed? it's been downgraded?

I assume this is from merging in master?

klawani: I think it happened because I did some local npm installation... Now upgraded to match master and ithc

config/default.json
```diff
@@ -106,7 +106,11 @@
       "citizen-oauth-client-secret": "123456",
       "os-postcode-lookup-api-key": "AAAAAAA",
       "encryptionKey": "not-a-real-32-bit-encryption-key",
-      "AppInsightsInstrumentationKey":"APPINSIGHTS_INSTRUMENTATIONKEY"
+      "AppInsightsInstrumentationKey":"APPINSIGHTS_INSTRUMENTATIONKEY",
+      "launchDarkly-sdk-key": "sdk-AAAAA"
     }
+  },
+  "launchDarkly": {
+    "offline": true
```
timja: this should default to false, it can be set to true in `development.json` and / or `mocha.json` if you want it to be enabled for tests

infrastructure/main.tf
```diff
@@ -149,6 +154,9 @@ module "citizen-frontend" {
 
     CONTACT_EMAIL = "${data.azurerm_key_vault_secret.staff_email.value}"
 
+    //LaunchDarkly
+    LD_SDK_KEY = "${data.azurerm_key_vault_secret.launch_darkly_sdk_key.value}"
+    LD_OFFLINE_FLAG = "${var.ld_offline_flag}}"
```
timja: this shouldn't need to be set as it will will default to on

config/default.json
klawani: done

infrastructure/main.tf
klawani: removed

infrastructure/main.tf
```diff
@@ -149,6 +154,8 @@ module "citizen-frontend" {
 
     CONTACT_EMAIL = "${data.azurerm_key_vault_secret.staff_email.value}"
 
+    //LaunchDarkly
```
timja: either 
```suggestion
    // LaunchDarkly
```
or remove this line, not sure if comment is needed

src/main/common/clients/launchDarklyClient.ts
```diff
@@ -0,0 +1,32 @@
...
+    if (!LaunchDarklyClient.client) {
+      LaunchDarklyClient.client = ld.init(sdkKeys, ldConfig)
+    }
+  }
+
+  async variation (user: User, roles: string[], featureKey: string, f?: () => void): Promise<ld.LDFlagValue> {
+    const ldUser: ld.LDUser = {
+      key: user.id,
+      custom: {
+        'roles': roles
```
timja: quotes shouldn't be needed here?
should be able to do:
```suggestion
        roles
```

timja:
any reason you merged this @klawani ? lint is failing on it

src/test/common/errorHandling.ts
```diff
@@ -25,18 +25,21 @@ describe('ErrorHandling', () => {
 
   it('should invoke given request handler', async () => {
     const handled = ErrorHandling.apply(resolvingRequestHandler)
+    // @ts-ignore
```
robertparkinson: why are these being added? would it not be better to fix the lint issue rather than ignore it?

src/test/features/claim/routes/pay.ts
```diff
@@ -438,26 +445,28 @@ describe('Claim issue: post payment callback receiver', () => {
         describe('when claim does not exist', () => {
 
           it('should return 500 and render error page when cannot save claim', async () => {
+            when(mockLaunchDarklyClient.variation(testUser, testRoles,'admissions')).thenResolve(Promise.resolve(false))
```
robertparkinson: can we get this to be done the same way as our other mocks for consistency?

PR #1606, opened 2019-04-30
ROC-5567 settlement dashboard tests

andrewwa-kainos:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-5567

### Change description
Add content route tests for settlement states on the dashboard and claim-status pages

- [ ] Yes
- [x] No


src/main/app/claims/models/claim.ts
```diff
@@ -91,9 +91,7 @@ export class Claim {
     return !this.countyCourtJudgmentRequestedAt
       && (this.admissionPayImmediatelyPastPaymentDate
         || this.hasDefendantNotSignedSettlementAgreementInTime()
-        || (!this.respondedAt && isPastDeadline(MomentFactory.currentDateTime(), this.responseDeadline)
-        || this.isSettlementAgreementRejected()
```
sabahirfan: The PR says only about adding tests.
This condition (`this.isSettlementAgreementRejected()`) is removed, @andrewwa-kainos I hope it's not accidental.

andrewwa-kainos: There was an understanding that any errors found would be fixed (hence some of the other changes). This change was more about cleaning up redundancy though.

sabahirfan: Thanks

src/main/features/dashboard/views/macro/caseHistory.njk
```diff
@@ -159,12 +159,12 @@
       {{ offerSubmittedForClaimantClaimDetails(claim) }}
     {% elif state.status === ClaimStatus.PART_ADMIT_PAY_IMMEDIATELY %}
       {{ partAdmittedPayImmediatelyClaimantClaimDetails(claim) }}
+    {% elif state.status === ClaimStatus.OFFER_SETTLEMENT_REACHED %}
```
simad: I'm so glad we're refactoring this `mess` :D

PR #1695, opened 2019-06-16
ROC-5810 fix for CCJ with payment date in the past

srineelesh:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-5810

### Change description
fix for CCJ with payment date in the past

### Work checklist

- [ X] Unit tests added where applicable
- [X ] Route tests added for new pages
- [ X] New pages included in a11y tests
- [X ] Task list and task completeness checks updated
- [ X] Check and send page updated
- [X ] Routes, page content and page flows of new features are toggled 
- [X ] UI changes look good on mobile
- [X ] Required Google Analytics events are being sent 

- [X ] I have clicked through the running application to see if all changes I made actually work.

- [X ] Yes
- [ ] No

### DIFF
--- a/src/main/app/claims/models/claim.ts
+++ b/src/main/app/claims/models/claim.ts
@@ -411,6 +411,23 @@ export class Claim {
       && this.settlement && this.settlement.isOfferRejected()
   }
 
+  get isSettlementPaymentDateValid (): boolean {
+    if (this.settlement) {
+      const offer = this.settlement.getLastOffer()
+      const now = MomentFactory.currentDate()
+      if (offer && offer.paymentIntention && offer.paymentIntention.paymentOption === PaymentOption.BY_SPECIFIED_DATE) {
+        const paymentDate = offer.paymentIntention.paymentDate
+        return (paymentDate.isAfter(now) || paymentDate.isSame(now))
+      } else if (offer && offer.paymentIntention && offer.paymentIntention.paymentOption === PaymentOption.INSTALMENTS) {
+        const firstPaymentDate = offer.paymentIntention.repaymentPlan.firstPaymentDate
+        return (firstPaymentDate.isAfter(now) || firstPaymentDate.isSame(now))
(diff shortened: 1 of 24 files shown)


src/main/app/claims/models/claim.ts
```diff
@@ -411,6 +411,23 @@ export class Claim {
       && this.settlement && this.settlement.isOfferRejected()
   }
 
+  get isSettlementPaymentDateValid (): boolean {
```
KiranVarma02: why is this a getter? i see that you followed above example but i think that's wrong too...if you see most other places in same file...they are all methods. 

I understand we didn't follow same standard everywhere in this file, i wish we did but i think we should do it right going forward..

KiranVarma02: looks like its wrongly called a getter in only 1 other place and you picked that one :)

simad: I agree with @KiranVarma02 , imho if something is computed on the fly with arguably complex logic it should be a method - leaving getters for returning properties with little manipulation. Ie. returning an uppercase string out of a property might as well be a getter

src/main/app/claims/models/claim.ts
```diff
@@ -411,6 +411,23 @@ export class Claim {
...
+    if (this.settlement) {
+      const offer = this.settlement.getLastOffer()
+      const now = MomentFactory.currentDate()
+      if (offer && offer.paymentIntention && offer.paymentIntention.paymentOption === PaymentOption.BY_SPECIFIED_DATE) {
+        const paymentDate = offer.paymentIntention.paymentDate
+        return (paymentDate.isAfter(now) || paymentDate.isSame(now))
+      } else if (offer && offer.paymentIntention && offer.paymentIntention.paymentOption === PaymentOption.INSTALMENTS) {
```
simad: all of these if conditions are based off `offer && offer.paymentIntention` so we might extract that in a separate condition and then switch over offer.paymentIntention.paymentOption. I think it would be way more readable

srineelesh: done

PR #1701, opened 2019-06-26
ROC-5818 Why do you disagree with the order page

robertparkinson:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-5818

### Change description

New page

### Work checklist

- [ ] Yes
- [x] No


src/main/features/orders/draft/ordersDraft.ts
```diff
@@ -0,0 +1,15 @@
+import { DraftDocument } from '@hmcts/cmc-draft-store-middleware'
+import { DisagreeReason } from 'orders/form/models/disagreeReason'
+
+export class OrdersDraft extends DraftDocument {
```
KiranVarma02: new Draft for just 1 property? I can see that it makes sense logically to keep it as separate draft but is it necessary?

robertparkinson: Where else would I save it? Or do you mean to not save the information at all? if the claim-store errors then it will be lost and someone could have put a lot of effort into writing their reason that would then be lost.

KiranVarma02: my understanding is orders come after DQ, so do you think we can just use existing DQ draft? just asking! i could be wrong.

robertparkinson: This story is for after the order has been drawn. it's after both parties have submitted their DQ and the legal advisor has drawn the order that this flow is made available.

src/main/features/orders/form/models/disagreeReason.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  deserialize (input: any): DisagreeReason {
+    if (input) {
+      this.reason = input.reason
+    }
+    return this
+  }
+
+  isCompleted (): boolean {
+    return !!this.reason && this.reason.length > 0
```
andrewwa-kainos: Should we be checking there's some non-whitespace?

src/main/features/orders/routes/disagree-with-order.ts
```diff
@@ -0,0 +1,52 @@
...
+      } else {
+
+        const draft: Draft<OrdersDraft> = res.locals.draft
+        const user: User = res.locals.user
+
+        draft.document.disagreeReason = form.model
+
+        await new DraftService().save(draft, user.bearerToken)
+
+        res.redirect(Paths.disagreeReasonPage.evaluateUri({ externalId: res.locals.claim.externalId }))
```
andrewwa-kainos: Would prefer to return to the dashboard until the next page is done

src/test/features/orders/form/models/disagreeReason.ts
```diff
@@ -0,0 +1,74 @@
...
+    it('should set the primitive fields to undefined', () => {
+      const reason = new DisagreeReason()
+      expect(reason.reason).to.be.undefined
+    })
+  })
+
+  describe('deserialize', () => {
+
+    it('should return an instance initialised with defaults for undefined', () => {
+      expect(new DisagreeReason().deserialize(undefined)).to.eql(new DisagreeReason())
```
andrewwa-kainos: Can we have more consistency with the language chains?

src/test/features/orders/form/models/disagreeReason.ts
```diff
@@ -0,0 +1,74 @@
...
+  })
+
+  describe('validation', () => {
+
+    const validator: Validator = new Validator()
+
+    it('should accept reason with empty string', () => {
+      const errors = validator.validateSync(new DisagreeReason(''))
+
+      expect(errors.length).to.equal(0)
```
andrewwa-kainos: Arrays and collections can use `.to.be.empty` and `.to.have.lengthOf(number)`

src/main/features/orders/form/models/disagreeReason.ts
robertparkinson: removed. there's no check-and-send page for this journey so no need for completed checks.

src/main/features/orders/routes/disagree-with-order.ts
robertparkinson: done

src/test/features/orders/form/models/disagreeReason.ts
robertparkinson: done

src/test/features/orders/form/models/disagreeReason.ts
robertparkinson: done

PR #1780, opened 2019-08-09
ROC-6175: Fix DQ validation issue

KiranVarma02:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-6175

### Change description
Fix DQ validation error

### DIFF
--- a/src/main/app/claims/models/directions-questionnaire/directionsQuestionnaire.ts
+++ b/src/main/app/claims/models/directions-questionnaire/directionsQuestionnaire.ts
@@ -50,8 +50,7 @@ export namespace DirectionsQuestionnaire {
         selfWitness: directionsQuestionnaire.selfWitness.option.option as YesNoOption,
         noOfOtherWitness: directionsQuestionnaire.otherWitnesses ? directionsQuestionnaire.otherWitnesses.howMany : undefined
       },
-      expertReports: (!(directionsQuestionnaire.expertReports.rows.length === 1
-        && directionsQuestionnaire.expertReports.rows[0].isEmpty())) ?
+      expertReports: (directionsQuestionnaire.expertReports && directionsQuestionnaire.expertReports.rows.length > 0) ?
         directionsQuestionnaire.expertReports.rows.map(row => ({
           expertName: row.expertName,
           expertReportDate: row.reportDate ? LocalDate.fromObject(row.reportDate).asString() : undefined
--- a/src/main/app/claims/models/directions-questionnaire/directionsQuestionnaire.ts
+++ b/src/main/app/claims/models/directions-questionnaire/directionsQuestionnaire.ts
@@ -50,19 +50,21 @@ export namespace DirectionsQuestionnaire {
         selfWitness: directionsQuestionnaire.selfWitness.option.option as YesNoOption,
         noOfOtherWitness: directionsQuestionnaire.otherWitnesses ? directionsQuestionnaire.otherWitnesses.howMany : undefined
       },
-      expertReports: directionsQuestionnaire.expertReports && directionsQuestionnaire.expertReports.rows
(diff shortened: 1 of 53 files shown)


src/test/app/claims/models/directions-questionnaire/directionsQuestionnaire.ts
```diff
@@ -56,6 +56,23 @@ describe('DirectionsQuestionnaire', () => {
       expect(DirectionsQuestionnaire.deserialize(directionsQuestionnaireDraftSampleData)).to.deep.equal(expectedData)
     })
 
+    it('should deserialize directions questionnaire correctly when no expert reports and no expert evidence', () => {
+      let directionsQuestionnaireDraftSampleData: DirectionsQuestionnaireDraft
```
robertparkinson: Change to const and investigate why you need the ts-ignore. can it be removed?

src/main/app/claims/models/directions-questionnaire/directionsQuestionnaire.ts
```diff
@@ -50,19 +50,21 @@ export namespace DirectionsQuestionnaire {
         selfWitness: directionsQuestionnaire.selfWitness.option.option as YesNoOption,
         noOfOtherWitness: directionsQuestionnaire.otherWitnesses ? directionsQuestionnaire.otherWitnesses.howMany : undefined
       },
-      expertReports: directionsQuestionnaire.expertReports && directionsQuestionnaire.expertReports.rows
-        && directionsQuestionnaire.expertReports.rows.map(row => ({
+      expertReports: (!(directionsQuestionnaire.expertReports.rows.length === 1
+        && directionsQuestionnaire.expertReports.rows[0].isEmpty())) ?
```
andrewwa-kainos: This smells like a hack. If possible, it would be better to ensure the `expertReports` is populated only with valid data when it's defined by the user.

KiranVarma02: i completely understand what you are saying... I thought the same. ExpertReports uses common code to deserializeRows which is used in other places in the app where ever we have multiple rows. Not sure why its originally developed in a way to generate empty but it does generate empty object when there are no rows. I decided not to change that thinking its probably like that for some valid reason. So, my other option is change how deserialise is done in expertRows, I could do that but thought this is okay!!

KiranVarma02: actually i am not sure why backend is like this 
` @Valid
    @Size(min = 1, max = 20)
    private final List<ExpertReport> expertReports;`

src/main/app/claims/models/directions-questionnaire/directionsQuestionnaire.ts
```diff
@@ -50,19 +50,20 @@ export namespace DirectionsQuestionnaire {
...
+        })) : undefined,
       unavailableDates: directionsQuestionnaire.availability &&
         directionsQuestionnaire.availability.unavailableDates.map(unavailableDate => ({
           unavailableDate: unavailableDate ? LocalDate.fromObject(unavailableDate).asString() : undefined
         })),
-      expertRequest: directionsQuestionnaire.expertEvidence && {
-        expertEvidenceToExamine: directionsQuestionnaire.expertEvidence.whatToExamine,
-        reasonForExpertAdvice: directionsQuestionnaire.whyExpertIsNeeded.explanation
-      }
+      expertRequest: (directionsQuestionnaire.expertEvidence.expertEvidence &&
```
simad: I find all of this logic very hard to read. We might want to extract it in a named functions for legibility

simad: Especially the extensive use of ternary operator plus that weird js `&&` "feature" makes me very sad

PR #1843, opened 2019-09-03
Enable mediation and dq in aat

KiranVarma02:
### JIRA link
No Jira

### Change description
Enable mediation and dq in aat


infrastructure/aat.tfvars
```diff
@@ -11,3 +11,5 @@ capacity = "2"
 feature_paid_in_full =  "true"
 feature_admissions = "true"
 feature_testing_support = "true"
+feature_mediation = "true"
```
markry-kainos: `feature_mediation_pilot = "true"` ??? we have this enabled in demo

KiranVarma02: done

PR #1846, opened 2019-09-04
Setup web chat

dharmendrak:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-6342
https://tools.hmcts.net/jira/browse/ROC-6341

### Change description

Setup web chat

- [ ] Yes
- [ ] No


src/main/app.ts
```diff
@@ -61,6 +61,8 @@ app.use(cookieEncrypter(config.get('secrets.cmc.encryptionKey'), {
   }
 }))
 
+// Web Chat
+app.use('/webchat', express.static(path.join(__dirname, '/node_modules/@hmcts/ctsc-web-chat/assets')))
```
robertparkinson: If you have copied everything into our folders in the above gulp file you shouldn't need to point to the node_modules folder. It should all be part of our files?

src/main/views/macro/support.njk
```diff
@@ -1,20 +1,41 @@
 {% from "externalLink.njk" import externalLink %}
 
+{% macro _webChatWidget(label = 'Web chat') %}
+      <h2 class="heading-medium" style="margin-bottom:5px; margin-top: 10px;">{{ t(label)}}</h2>
+      <div id="ctsc-web-chat"></div>
+{% endmacro%}
+
+{% macro _emailWidget(email = 'moneyclaims@justice.gov.uk') %}
```
robertparkinson: this email should be retrieved from the CONTACT_EMAIL variable

src/main/app.ts
sabahirfan: Shouldn't we use from the library itself rather than copying locally? We don't want to end up different versions in our local.

robertparkinson: every time you do a yarn install then do a gulp it should copy and over-ride the files to the correct folders. We may need to add in a script to clear out the old ones. You normally only get problems if they delete files from their repo and we don't clear it up with gulp

src/main/views/macro/support.njk
dharmendrak: done

src/main/app.ts
sabahirfan: 👍

PR #1866, opened 2019-09-11
ROC-6452 Only show intention to proceed to new claims

robertparkinson:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-6452

### Work checklist

- [ ] Yes
- [ ] No


src/main/app/claims/models/claim.ts
```diff
@@ -646,4 +646,8 @@ export class Claim {
   private hasOrderBeenDrawn (): boolean {
     return !!this.directionOrder
   }
+
+  public isIntentionToProceedEligible (): boolean {
+    return this.directionsQuestionnaireDeadline.isAfter(MomentFactory.parse('2019-09-29'))
```
KiranVarma02: 2019-09-29??

KiranVarma02: Do you need method to be public?

src/main/app/claims/models/claim.ts
```diff
@@ -646,4 +646,8 @@ export class Claim {
   private hasOrderBeenDrawn (): boolean {
     return !!this.directionOrder
   }
+
+  public isIntentionToProceedEligible (): boolean {
```
KiranVarma02: Do you need method to be public?

src/main/app/claims/models/claim.ts
KiranVarma02: if we need this date for some reason- it may be useful to name it as a constant with some name explaining why we use this date.

src/main/app/claims/models/claim.ts
robertparkinson: it is being called from the nunjucks file. I was just emulating one of the other calls

src/main/app/claims/models/claim.ts
robertparkinson: we set the date of the deadline to be 19 days in the future. so a claim that gets a full defence on monday will have a deadline date set of the 30th

PR #1903, opened 2019-10-11
GitHub actions trial

robertparkinson:
### Work checklist

- [ ] Yes
- [ ] No


package.json
```diff
@@ -9,14 +9,16 @@
     "setup": "gulp sass copy-files",
     "start": "node server.js",
     "lint": "tsc --noEmit && tslint --project tsconfig.json && sass-lint -v -q",
-    "test": "TS_NODE_TRANSPILE_ONLY=true NODE_ENV=mocha LOG_LEVEL=OFF mocha --opts mocha.opts $(find src/test \\( -name '*.ts' \\) ! -path '*a11y*' ! -path '*/routes/*') --reporter-options reportFilename=unit,inlineAssets=true,reportTitle=citizen-frontend-unit",
+    "test": "echo 'unit tests are ran through GitHub actions as a mandatory step in order to save time on the main pipeline'",
```
KiranVarma02: Did you miss actual test here - ```TS_NODE_TRANSPILE_ONLY=true NODE_ENV=mocha LOG_LEVEL=OFF mocha --opts mocha.opts $(find src/test \\( -name '*.ts' \\) ! -path '*a11y*' ! -path '*/routes/*') --reporter-options reportFilename=unit,inlineAssets=true,reportTitle=citizen-frontend-unit"```

robertparkinson: The test:coverage runs all the unit and route tests so no need to duplicate them

KiranVarma02: there is no test:coverage after your changes isn't it? do you mean "test:action-coverage"? I would prefer test to run unit tests, coverage to run unit tests is misleading i think..

robertparkinson: the test coverage has been moved to the github action yes (test:action-coverage). we need to run the coverage test and as part of that test it runs the unit and route tests.

KiranVarma02: I don't like the idea of running coverage to run unit and route tests unless there is some valid reason. I see that PR is still WIP, it's probably something you can think about ..

robertparkinson: I've now created tests and tests:a11y so it is still similar to how we normally do it

PR #1905, opened 2019-10-14
ROC-4043 claimant provide defendant phone

karthikkannankainos:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-4043

### Change description
Claimant provides the defendant phone number.


src/main/app/claims/claimModelConverter.ts
```diff
@@ -246,4 +251,11 @@ export class ClaimModelConverter {
       date_created: payment.date_created
     }
   }
+
+  private static convertPhoneNumber(phone: MobilePhone): string {
+    if (phone === undefined) {
+      return undefined;
+    }
+    return StringUtils.trimToUndefined(phone.number)
+  }
```
andrewwa-kainos: Doesn't cover `null`, which is also falsy. How about just
`return phone ? StringUtils.trimToUndefined(phone.number) : phone`
then
`undefined` => `undefined`
`null` => `null`
`{"number":"something"}` => `"something"`
`{}` => `undefined`
`{"number":"     "}` => `undefined`

src/main/features/claim/views/check-and-send.njk
```diff
@@ -42,6 +42,9 @@
     }}
 
     {{ row('Email', draftClaim.defendant.email.address, paths.defendantEmailPage.uri, bold = true) }}
+    {% if draftClaim.defendant.mobilePhone %}
+      {{ row('Contact number (optional)', draftClaim.defendant.mobilePhone.number, paths.defendantPhonePage.uri, bold = true) }}
+    {% endif %}
```
andrewwa-kainos: It should still be shown in the check-and-send page if it's absent, so the user sees that it hasn't been entered and gets the link to go and change it.

src/main/app/claims/claimModelConverter.ts
karthikkannankainos: Done. Unable to send phone back as type clash. so sending undefined back.

src/main/features/claim/views/check-and-send.njk
karthikkannankainos: sorted it.

PR #1908, opened 2019-10-16
ROC-6000 Let user type in commas according to currency format

Valkrys:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-6000

### Change description
User can now use commas inline with currency formatting and the total is added up correctly. If the format is incorrect, the total does not add up and upon submit, errors appear according to the pre-existing validation.

### Work checklist

- [ ] Yes
- [x] No


src/main/public/js/calculate-amount-total.js
```diff
@@ -12,15 +13,20 @@ $(document).ready(function () {
     return true
   }
 
-  $('input.claim-amount-column-2').keyup(updateTotal)
+  $('#claim-amount').on('keyup','input.claim-amount-column-2', updateTotal)
 
   function toNumber (index, inputElement) {
-    var parsed = parseFloat($(inputElement).val())
+    var regex =  /^(\d+|\d{1,3}(,\d{3})*)(\.\d+)?$/;
```
andrewwa-kainos: We validate that the input has a maximum of 2dp. We could put that in this regex too
`... (\.\d\d?)?` or `... (\.\d{1,2})?`

PR #2014, opened 2019-11-29
ROC-7043 AppInsights custom event for all rendering errors

andrewwa-kainos:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-7043

### Change description
- add configuration to log app insights events to the local console
- send app insights event for all template rendering errors

- [ ] Yes
- [x] No


src/main/modules/app-insights/index.ts
```diff
@@ -1,35 +1,24 @@
...
-        if (envelope.tags['ai.operation.name']) {
-          envelope.tags['ai.operation.name'] = hideUuidInUrlIfNotStaticFile(envelope.tags['ai.operation.name'])
-        }
-      }
+    appInsights.defaultClient.addTelemetryProcessor(telemetryProcessors.operationNameUUIDHider())
 
-      // always send
-      return true
-    })
+    if (aiInstrumentationKey === 'STDOUT') {
```
daveoh: Should we have a test around this branch?

andrewwa-kainos: There were previously zero tests for this file; I refactored the telemetry bits out purely for testability. Not sure how I would go about testing this in TypeScript. If you know how, please share! As it stands, it's already improved coverage

andrewwa-kainos: Have tried using `mock-require` but couldn't get it working. I don't feel it's worth spending hours trying to unit test this.

PR #2165, opened 2020-02-05
ROC-7263: Incorrect fees

KiranVarma02:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-7262

### Change description

Incorrect fees


src/test/http-mocks/fees.ts
```diff
@@ -66,6 +66,228 @@ export const feeRange = [
...
+      'memo_line': 'RECEIPT OF FEES - Civil issue money',
+      'statutory_instrument': '2015 No 576',
+      'si_ref_id': '1.2b',
+      'natural_account_code': '4481102133',
+      'fee_order_name': 'Civil Proceedings',
+      'direction': 'enhanced'
+    }
+  },
+  {
+    'code': 'FEE0212',
```
pragnesh-patel: We have in our mocks Fee Code hard-coded - I'm assuming we will need to change it for newer fee codes generated & different versions of API?

KiranVarma02: that's just test data.

PR #2212, opened 2020-03-04
ROC-7622 AppInsights traces when automatically redirecting to the log-in page

andrewwa-kainos:
# JIRA link
https://tools.hmcts.net/jira/browse/ROC-7622

### Change description
* Add new custom event tracker method for traces
  * which passes up the message to AppInsights
* Use the new tracer method in the authorisation middleware instead of just logging to the console
* Add new telemetry processor which logs app insights messages to console when the API key is `STDOUT`

- [ ] Yes
- [x] No


src/main/app/idam/authorizationMiddleware.ts
```diff
@@ -37,14 +38,14 @@ export class AuthorizationMiddleware {
       }
 
       if (!jwt) {
-        logger.debug(`Protected path - no JWT - access to ${req.path} rejected`)
+        CustomEventTracker.trackTrace(`Protected path - no JWT - access to ${req.path} rejected`)
```
timja: this should get logged to app insights if you log at info level

pradhap87:
There is no active commits in this branch, close to free up the space in pipeline.

PR #2226, opened 2020-03-12
Rename web chat config variables names

dharmendrak:


src/main/routes/webchat.ts
```diff
@@ -1,14 +1,24 @@
 import * as express from 'express'
 import * as config from 'config'
 
-class Paths {
+class WebChat {
   static main: string = '/webchat'
+
+  static filterSecrets (accepted, secrets) {
```
sabahirfan: Please add unit test for this method.

dharmendrak: added

PR #2294, opened 2020-04-23
Amend preview chart for redirect config

dharmendrak:


charts/cmc-citizen-frontend/values.preview.template.yaml
```diff
@@ -1,5 +1,5 @@
 tags:
-  ccd-idam-pr: true
```
klawani, line 2: I think we need to remove this completely... it's not needed anymore. It's already enabled in the chart we are using.
https://github.com/hmcts/chart-ccd/blob/master/ccd/values.yaml#L2

charts/cmc-citizen-frontend/values.preview.template.yaml
```diff
@@ -151,6 +151,8 @@ ccd:
       keyVaults:
       environment:
         IDAM_OAUTH2_AW_CLIENT_SECRET: ${ADMIN_WEB_IDAM_SECRET}
+    idam-pr:
```
klawani, line 154: you don't really need this. it's already in ccd-chart-4 we are using

charts/cmc-citizen-frontend/values.preview.template.yaml
dharmendrak: removed

charts/cmc-citizen-frontend/values.preview.template.yaml
dharmendrak: I think it is needed

PR #2328, opened 2020-05-21
ROC-7534 Allow online journey after offline defence

Valkrys:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-7534

### Change description
Creating new dashboard that will allow defendant and claimant to continue after an offline response.

- [ ] Yes
- [x] No


src/main/features/dashboard/views/macro/caseStatus.njk
```diff
@@ -124,6 +124,8 @@
     {{ ccjRequestedForClaimantDashboard(claim) }}
   {% elif status === ClaimStatus.REDETERMINATION_BY_JUDGE %}
     {{ redeterminationClaimantDashboard() }}
+  {% elif claim.response.responseMethod == ResponseMethod.OCON_FORM %}
```
daveoh, line 127: `ClaimStatus` Is a status that is only used on the frontend (i.e it holds more states that ClaimState on the ClaimStore) and is built in `claim.ts`

We probably should have a new status of `DEFENDANT_OFCOM_RESPONSE` and then follow the pattern of checking this status in this god awful nested if structure. 

This way it would be easier to refactor this if statement to a switch instead which is something that needs to be done in the future.

src/main/features/dashboard/views/macro/caseStatus.njk
```diff
@@ -124,6 +124,8 @@
     {{ ccjRequestedForClaimantDashboard(claim) }}
   {% elif status === ClaimStatus.REDETERMINATION_BY_JUDGE %}
     {{ redeterminationClaimantDashboard() }}
+  {% elif claim.status === ClaimStatus.DEFENDANT_OCON_OFFLINE_RESPONSE %}
```
daveoh, line 127: Can just be
```suggestion
  {% elif status === ClaimStatus.DEFENDANT_OCON_OFFLINE_RESPONSE %}
```

src/main/features/dashboard/views/macro/caseStatus.njk
```diff
@@ -236,6 +238,8 @@
     {{ redeterminationDefendantDashboard(claim) }}
   {% elif status === ClaimStatus.CCJ_REQUESTED %}
     {{ ccjRequestedForDefendantDashboard(claim) }}
+  {% elif claim.status === ClaimStatus.DEFENDANT_OCON_OFFLINE_RESPONSE%}
```
daveoh, line 241: Same here
```suggestion
  {% elif status === ClaimStatus.DEFENDANT_OCON_OFFLINE_RESPONSE%}
```

andrewwa-kainos:
Where/when is responseMethod written to the data? How are you handling existing data - will you do a manual DB update to inject the data? If so, how will you work out which ones it applies to?

daveoh:
> Where/when is responseMethod written to the data? How are you handling existing data - will you do a manual DB update to inject the field? If so, how will you work out which ones it applies to?
@andrewwa-kainos This is handled in this PR:
https://github.com/hmcts/cmc-claim-store/pull/1709/files

sabahirfan:
Along with the questions raised by @andrewwa-kainos, don't we need any tests for this change?

PR #2329, opened 2020-05-27
Features/ROC-7651 turn on admissions for all new claims

sudheer-ajax:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-7651

### Work checklist

- [ ] Yes
- [ ] No


src/main/features/response/helpers/taskListBuilder.ts
```diff
@@ -41,11 +41,6 @@ export class TaskListBuilder {
     const tasks: TaskListItem[] = []
     const externalId: string = claim.externalId
 
-    if (!ClaimFeatureToggles.isFeatureEnabledOnClaim(claim)
```
tharacka, line 44: Should the test for ` (draft.isResponsePartiallyAdmitted() || draft.isResponseFullyAdmitted())` remain?

src/test/features/dashboard/routes/claimStatus/full-defence.ts
```diff
@@ -702,8 +702,8 @@ const mediationDQEnabledClaimDetails = [
     status: 'Full defence - defendant dispute all of the claim and rejects mediation with directions questionnaire enabled',
     claim: fullDefenceClaim(),
     claimOverride: {
-      ...directionsQuestionnaireDeadline(),
```
tharacka, line 705: will remove `()` cause an issue?

src/main/features/response/helpers/taskListBuilder.ts
sudheer-ajax: No because this condition applies only if there is no admissions feature in claim and if you see response type PA of FA. After making admission as a default feature we don't

sudheer-ajax: No because this condition applies only if there is no admissions feature in claim and if you see response type PA of FA. After making admission as a default feature we don't need to check response type as it is allowed always and no need to delete from draft

src/test/features/dashboard/routes/claimStatus/full-defence.ts
sudheer-ajax: Fixed

PR #2335, opened 2020-06-03
Feature/roc 7977 transferred case status

daveoh:
### JIRA link
https://tools.hmcts.net/jira/browse/ROC-7977

### Change description
Transferred case status

### Work checklist

- [ ] Yes
- [x] No


src/test/app/claims/models/claim.ts
```diff
@@ -1044,6 +1044,19 @@ describe('Claim', () => {
     })
   })
 
+  describe('isTransferred', () => {
+    let claim
+
+    beforeEach(() => {
+      claim = new Claim()
+      claim.state = 'TRANSFERRED'
+      claim.responseDeadline = MomentFactory.currentDate()
```
justin-r-oliver, line 1053: I don't think we need this line

daveoh: This is needed as some of the calculations to determine what the status is depend on a value in this variable

PR #2339, opened 2020-06-05
ROC-7807 view scanned defence response

justin-r-oliver:
### JIRA link

https://tools.hmcts.net/jira/browse/ROC-7807

### Change description

Allow claimants to view a defendants scanned response form (OCON9x)

- [ ] Yes
- [X] No


src/main/app/documents/scannedDocumentsClient.ts
```diff
@@ -0,0 +1,44 @@
...
+      headers: {
+        Authorization: `Bearer ${bearerToken}`,
+        Accept: 'application/pdf'
+      },
+      encoding: null
+    }
+
+    return this.request(options).then(function (response) {
+      return response
+    })
```
andrewwa-kainos, line 42: No point transforming the result into itself
```suggestion
    return this.request(options)
```

src/test/app/documents/scannedDocumentsClient.ts
```diff
@@ -0,0 +1,49 @@
...
+  it('should should retrieve the scanned response form for a case', () => {
+
+    const expectedResponse = {
+      url: 'http://localhost/scanneddoc'
+    }
+
+    nock(mockClient)
+      .get(/\/scanned-documents\/.+/)
+      .reply(200, expectedResponse)
+    return client.getScannedResponseFormPDF(externalId, bearerToken)
```
andrewwa-kainos, line 25: Need to verify the details of the result, not return it

andrewwa-kainos:
Please correct the PR title. It should be `ROC-7807 view scanned defence response` in this case, if you don't amend the textual content.

src/main/features/response/routes/scanned-response-form.ts
```diff
@@ -0,0 +1,19 @@
...
+import { DownloadUtils } from 'utils/downloadUtils'
+
+const scannedDocumentsClient: ScannedDocumentsClient = new ScannedDocumentsClient()
+
+/* tslint:disable:no-default-export */
+export default express.Router()
+  .get(Paths.scannedResponseForm.uri,
+    ErrorHandling.apply(async (req: express.Request, res: express.Response, next: express.NextFunction) => {
+      const { externalId } = req.params
+      const pdf: Buffer = await scannedDocumentsClient.getScannedResponseFormPDF(externalId, res.locals.user.bearerToken)
```
andrewwa-kainos, line 15: I would suggest this controller shouldn't know the ScannedDocumentsClient needs a bearer token specifically, but needs to get the response form for the given claim (external ID) with that _user_'s authorisation. If you agree, refactor the method to take the `User` instead of the `bearerToken` and then the client will know that what it needs from the user is its bearer token.

PR #2448, opened 2020-08-06
Citizen frontend Accessibility (DAC) Issue Fix

pradhap87:
### JIRA link
Citizen frontend Accessibility (DAC) Issue Fix

### Change description

https://tools.hmcts.net/jira/browse/ROC-7787

### Work checklist

- [ ] Yes
- [x] No


src/test/features/claim/routes/check-and-send.ts
```diff
@@ -85,10 +85,10 @@ describe('Claim issue: check and send page', () => {
           .get(ClaimPaths.checkAndSendPage.uri)
```
prudhvi-v, line 85: Wouldn't it be good to have '<a href='.....>' in a variable and use it in multiple places.?

pradhap87: we have different types of <a href= ....> It will be nice to leave it AS-IS instead of them into variable. approximately we need to create 8 different variables to 

image

image

pradhap87: image

src/main/views/macro/table.njk
```diff
@@ -20,7 +20,12 @@
     </div>
     {% if changeLink %}
       <div class="column-one-quarter change-link-column">
-        <a href="{{ changeLink | safe }}" class="bold">{{ t('Change') }} <span class="visuallyhidden">{{ label | lower }}</span></a>
+        {% if (label === 'Full name') %}
+          <a href="{{ changeLink | safe }}" class="bold">{{ t('Change') }} <span class="visuallyhidden">{{ 'your full name, address and correspondence address' | lower }}</span></a>
```
prudhvi-v, line 24: This is not just one place we need to update. Noticed there are other sections that need similar fix on the same page. Below are few details where I noticed the issue.

- Claimant > check and submit > Their Details (this also has similar bug)
-  Claimant > check and submit > Claimant Amount
- Defendant > journey > YET TO BE LOOKED INTO
- For Defendant we need to check for 'Individual', 'Company', .... etc as well

pradhap87: It's not a good idea to refactor this in DAC ticket, we can have a bug raised and take this one. I will create a bug for this one as this is an existing code.

pradhap87: Created the bug ROC-8242 for this issue.

JIRA URL : https://tools.hmcts.net/jira/browse/ROC-8242

PR #2461, opened 2020-08-11
ROC-8226 Show other party details

singhtanveer:
Initial Code - Show other party details(Name, Address, Email & Phone) to Claimant and Defendant

### JIRA link
https://tools.hmcts.net/jira/browse/ROC-8226

### Work checklist

- [ ] Yes
- [ ] No


src/main/features/dashboard/index.ts
```diff
@@ -34,6 +35,28 @@ function render (claim: Claim, type: string): string {
   }
 }
 
+function formatAddress (address: Address): string {
```
prudhvi-v, line 38: could have written using 'for in' or Object.keys

src/main/features/dashboard/views/macro/partyDetails.njk
```diff
@@ -0,0 +1,29 @@
...
+  <h2 class="heading-medium no-top-margin">{{ otherParty }}'s details</h2>
+  <table class="govuk-table">
+    <tbody class="govuk-table__body">
+      <tr class="govuk-table__row">
+        <td class="govuk-table__cell--left">Name</td>
+        <td class="govuk-table__cell--right">{{ partyClaimData.name }}</td>
+      </tr>
+      <tr class="govuk-table__row">
+        <td class="govuk-table__cell--left">Address</td>
+        <td class="govuk-table__cell--right">{{ partyClaimData.address | formatAddress | default('Not available') | safe }}</td>
```
prudhvi-v, line 12: any specific reason for using 'safe' here?. this is needed only if absolutely needed

src/main/features/dashboard/index.ts
singhtanveer: if the sequence of key changes for some reason, address may start with zip code or city name instead of street name. 
also if any new key gets introduced which ideally should not, it may cause problem. this way we have better control, I think.

src/main/features/dashboard/views/macro/partyDetails.njk
singhtanveer: if safe is not used, data appears as simple text showing HTML code, safe renders code as DOM elements rather than plain text

pradhap87 (approved):

ThirumuruganD (approved):

PR #3383, opened 2022-05-11
Civ 490 keywords using event type and amount

darrenHMCTS:
### JIRA link
https://tools.hmcts.net/jira/browse/CIV-490

### Work checklist

- [ ] Yes
- [ ] No

### DIFF
--- a/src/main/app/utils/keywordHearingEventUtils.ts
+++ b/src/main/app/utils/keywordHearingEventUtils.ts
@@ -1,0 +1,23 @@
+import { listHearingKeywords } from 'listHearingKeywords';
+
+export class KeywordHearingEventUtils {
+  static getKeywordHearingEvent (amount: number) {
+    let keyword: String = '';
+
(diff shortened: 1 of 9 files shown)


src/main/app/utils/keywordHearingEventUtils.ts
```diff
@@ -0,0 +1,23 @@
+import { listHearingKeywords } from './listHearingKeywords'
+
+export class KeywordHearingEventUtils {
+  static getKeywordHearingEvent (amount: number) {
+    let keyword: string = ''
+
+    if (amount >= 0.01 && amount <= 300) {
+      keyword = listHearingKeywords[0]
```
sabahirfan, line 8: relying on the index of array does not seems to be nice here, as if we add/remove anything in array this while logic will break. Let's use constants/enum for each instead

sabahirfan (approved):

PR #3609, opened 2023-07-19
update javascript

zikrurR:
### Work checklist

- [ ] Yes
- [ ] No

### DIFF
--- a/src/main/public/js/analytics-custom-event-sender.js
+++ b/src/main/public/js/analytics-custom-event-sender.js
@@ -67,7 +67,10 @@ $(function () {
   // Send a google analytics event when an element that has the 'analytics-click-event-trigger' class is clicked.
   $('.analytics-click-event-trigger').on('click', function () {
     var cookies_policy = getCookie("cookies_policy");
-    var json = JSON.parse(this.response);
+    var json = null;
+    if (typeof this.reponse !== 'undefined' || this.response !== undefined){
+      json = JSON.parse(this.response);
+    }
     if (!cookies_policy && cookies_policy.split(',')[1].split(':')[1] === 'true')
     {
       var label = $(this).data('eventLabel')
--- a/src/main/public/js/analytics.js
+++ b/src/main/public/js/analytics.js
@@ -13,7 +13,8 @@ xhttp.onreadystatechange = function() {
     })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
 
     var cookiesPolicy = getCookie("cookies_policy");
-    if (cookiesPolicy.split(',')[1].split(':')[1] === 'true') {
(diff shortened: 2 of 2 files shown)


src/main/public/js/analytics.js
```diff
@@ -13,7 +13,8 @@ xhttp.onreadystatechange = function() {
     })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
 
     var cookiesPolicy = getCookie("cookies_policy");
-    if (cookiesPolicy.split(',')[1].split(':')[1] === 'true') {
+
+    if (cookiesPolicy !== '' && (cookiesPolicy.split(',')[1].split(':')[1] === 'true')) {
```
sabahirfan, line 17:
```suggestion
    if (cookiesPolicy !== undefined && (cookiesPolicy.split(',')[1].split(':')[1] === 'true')) {
```

src/main/public/js/analytics-custom-event-sender.js
```diff
@@ -67,7 +67,10 @@ $(function () {
   // Send a google analytics event when an element that has the 'analytics-click-event-trigger' class is clicked.
   $('.analytics-click-event-trigger').on('click', function () {
     var cookies_policy = getCookie("cookies_policy");
-    var json = JSON.parse(this.response);
+    var json = null;
+    if (typeof this.reponse !== 'undefined' || this.response !== undefined){
```
sabahirfan, line 71:
```suggestion
    if (this.reponse !== null && this.response !== undefined){
```

sabahirfan (approved):

src/main/public/js/analytics-custom-event-sender.js
dtortolaV1, line 71: this.reponse instead of this.response
Also, JSON.parse(null) does not fail? why not use if (this.response) json = JSON.parse(this.response)

dtortolaV1: And also, uses of json from this point on need null check, like lines 78, 80 when json.gaTrackingId is called

src/main/public/js/analytics.js
```diff
@@ -13,7 +13,8 @@ xhttp.onreadystatechange = function() {
     })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
 
     var cookiesPolicy = getCookie("cookies_policy");
-    if (cookiesPolicy.split(',')[1].split(':')[1] === 'true') {
+
```
dtortolaV1, line 16: isn't this code in the same case that the file above? I mean with the risk of null/undefined this.response

zikrurR: I'm not sure what you are saying here, however the premise for this to occur has changed and this would only resolve if the conditions for the premise has been satisfied.

asthamalviya (approved):

